Issue #393, opened 2021-11-16
Timestamp fields returned as JavaScript Dates?


enjalot:
It would be great if the client could convert timestamp fields to native JavaScript Date objects instead of the millisecond value.

Here is an example using both a parquet file and a csv file articulating the ask:
https://observablehq.com/@observablehq/duckdb-date-types

enjalot:
It looks like Timestamp fields in the SQL database are returned as Timestamp fields in Arrow. There is a Date field in Arrow that does return native JavaScript dates. I haven't been able to figure out how to cast from Timestamp to Date, but it seems like it would make sense to be able to do that?

ankoh:
Yes, DuckDB emits Arrow Timestamps in these cases. It is the more general type that can be specified with multiple units and an optional timezone.
DuckDB does not have support for time zones but this is [being worked on](https://github.com/duckdb/duckdb/issues/561).

Arrow does have the type Date64 which sits in a weird spot as it is defined as ms since epoch.
So Date64 seems to be a timestamp with millisecond resolution and no attached time zone? If that's the case we could introduce a flag in DuckDB-Wasm to autocast these timestamps to Date64.

Coming back to js-dates: I assume Arrow cannot emit plain js dates for timestamps by default since the Arrow timestamp type is too flexible. If, however, arrow-js maps Date64 to DateVectors by default, we can maybe live with such an autocast.

domoritz:
Arrow has multiple date types so one of them may be the right fit. The date types are built from and return JS `Date` instances. The timestamp types are built and return numbers/bigints.

ankoh:
Arrow does not have a plain Date type. Only in Js.
They have a 32 bit date32 that encodes days and a 64 bit date64 that encodes milliseconds since Unix epoch.
https://arrow.apache.org/docs/python/api/datatypes.html
The key question is whether it's semantically sound to autocast timestamp(ms) to date64 or whether this will turn into a footgun later.

enjalot:
Is there someway I could prepare some tests or examples?
I was thinking of creating a parquet file using python to create columns for all the various date/time types with a few consistent dates. then could look at the resulting JS arrow and native JS interpretations?

ankoh:
We could test this fairly easily without adding a parquet file right? I think there's some value in heaving self-contained jasmine tests if that's possible.
So in DuckDB-Wasm we have a fair amount of C++ tests and a(/too) few Typescript tests.
[...]

PR #401, opened 2021-11-18
add failing tests for #393

enjalot:
I've attempted to create a test that would pass if the desired behavior in #393 is implemented.


ankoh:
i'll build the autocast against this test, thx a lot.

ankoh:
This is now merged, thanks!

PR #417, opened 2021-11-22
Update README to use Uint8Array

jimmyhmiller:
`registerFileBuffer` expects a`Uint8Array`, not an `ArrayBuffer`. This updates the parquet example in the readme to properly use that type.

See:
https://github.com/duckdb/duckdb-wasm/blob/84ead1066ee716731e30420dcae81f72e1fc7e7b/packages/duckdb-wasm/src/bindings/bindings_interface.ts#L31


ankoh:
Thanks!

PR #476, opened 2021-12-17
feat: add castDurationToTime64 option

### LINKED ISSUE #470: Interval type unsupported in query output
Interval Type is not working as expected.

Query `SELECT INTERVAL '3' MONTH AS interval`
- on shell.duckdb.org: outputs an empty line.
- on DuckDB cli it returns: 

```
┌──────────┐
│ interval │
├──────────┤
│ 3 months │
└──────────┘
```

Reason for this appears to be that DuckDB sets the type for Interval to the Duration type ("tdm" type from [arrow c data interface docs](https://arrow.apache.org/docs/format/CDataInterface.html?highlight=tDm)). The Interval is converted to milliseconds in DuckDB. The duration type that is emitted by DuckDB, is not supported in Arrow JS. 

~The most likely option seems to be for DuckDB to emit an Interval type that JS does support. There seems to be some support for Intervals as the type is listed in the docs. If DuckDB emits interval [days, time] for the interval type DuckDB-WASM would be able to support intervals.~

(edit) Just discussed the possibility of emitting arrow interval types from DuckDB with mark, the problem is that the interval types of DuckDB and and arrow are not compatible. The best solution now seems to wait for the Duration type to be implemented in Arrow JS.

samansmink:
Fixes #470 by adding a castDurationToTime64. Note that its turned on by default and will cast the value emitted by DuckDB from milliseconds to microseconds.

### DIFF
--- a/lib/include/duckdb/web/config.h
+++ b/lib/include/duckdb/web/config.h
@@ -46,7 +46,10 @@ struct QueryConfig {
     std::optional<bool> cast_duration_to_time64 = std::nullopt;
 
     /// Has any cast?
-    bool hasAnyCast() const { return cast_bigint_to_double.value_or(false) || cast_timestamp_to_date.value_or(false) || cast_duration_to_time64.value_or(false); }
+    bool hasAnyCast() const {
+        return cast_bigint_to_double.value_or(false) || cast_timestamp_to_date.value_or(false) ||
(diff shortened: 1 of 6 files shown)


ankoh (approved):

ankoh:
Awesome, thx!

Issue #477, opened 2021-12-17
cast decimal to double option


samansmink:
Just found out that Arrow JS has a open issue where negative decimal types do not work properly ([link](https://issues.apache.org/jira/browse/ARROW-6575))

Maybe its good idea to add a fourth cast option similarly to PR #476 called `castDecimalToDouble`? I can do another PR for it?

ankoh:
Sure, sounds good! We can just add a deprecation warning later when they are resolved. 👍🏻

bluehat974:
Weird negative number will be solved in apache-arrow 13
https://github.com/apache/arrow/pull/35067

PR #487, opened 2021-12-21
Test/all types test

samansmink:
Integrated the test_all_types() function that was added to duckdb in https://github.com/duckdb/duckdb/pull/2739 to assert all types are working in duckdb-wasm. It will help keep track of the types of DuckDB that are supported. Note that the test will break if the DuckDB version is updated with new types. 

For the timezone types such as Date_tz, these are emitted by the test_all_types function, but because DuckDB does not yet fully support these I can't easily test custom values as their limits are invalid JS dates.

I'll update the decimals in this test as soon as I fix https://github.com/duckdb/duckdb-wasm/issues/477

### DIFF
--- a/packages/duckdb-wasm/test/all_types.test.ts
+++ b/packages/duckdb-wasm/test/all_types.test.ts
@@ -1,17 +1,22 @@
 import * as duckdb from "../src/";
 import {Column, Vector} from "apache-arrow";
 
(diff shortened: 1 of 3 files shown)


ankoh:
Awesome, thx!
Could you please rebase 47d02c4 onto the latest master and remove f6b7995?
I'd love to preserve the linear history.

PR #490, opened 2021-12-22
Feature/cast decimal to float

samansmink:
The castDecimalToDouble function that be used as a workaround for https://github.com/duckdb/duckdb-wasm/issues/477. 

The included tests should break if the issue is solved in arrowjs so that we know when we can deprecate this.


ankoh (approved):
lgtm

PR #518, opened 2022-01-04
Test/all types test cpp

samansmink:
Wrote the all_types_test on the cpp side as well, similar to the ts test from #487


.vscode/settings.json
```diff
@@ -1,75 +0,0 @@
-{
```
ankoh, line 1: can we keep this file here?

ankoh:
Looks great, thx! :)

PR #557, opened 2022-01-19
Feature/s3 fs

samansmink:
hey @ankoh here's the s3fs for reading. I've done my best to write enough meaningful test without requiring any mocking or s3 server emulating, but there's tests that really only work with an s3 server. I think for now its okay though and in line with DuckDB, we could later look into mocking s3 or using something like [s3rver](https://www.npmjs.com/package/s3rver) to make the tests more solid.

### DIFF
--- a/packages/duckdb-wasm/src/utils/s3_helper.ts
+++ b/packages/duckdb-wasm/src/utils/s3_helper.ts
@@ -1,0 +1,215 @@
+import {S3Config} from "../bindings";
+import {sha256} from "js-sha256";
+
+interface S3Params {
+    url: string,
+    query: string,
+    host: string,
+    region: string,
+    service: string,
+    method: string,
+    access_key_id: string,
+    secret_access_key: string,
+    session_token: string,
+    date_now: string,
+    datetime_now: string
+}
+
+const getS3Params = function(config : S3Config | undefined, url: string, method : string) : S3Params {
+    const parsedS3Url = parseS3Url(url);
+    return {
+        url: parsedS3Url.path,
+        query: "",
+        host: `${parsedS3Url.bucket}.s3.amazonaws.com`,
+        region: (config?.region) ?? "",
+        service: "s3",
+        method: method,
+        access_key_id: (config?.accessKeyId) ?? "",
+        secret_access_key: (config?.secretAccessKey) ?? "",
+        session_token: (config?.sessionToken) ?? "",
+        date_now: new Date().toISOString().replace(/-/g,'').split('T')[0],
+        datetime_now: new Date().toISOString().replace(/-/g,'').replace(/:/g,'').split('.')[0]+ 'Z',
+    };
+};
+
+const uri_encode = function(input : string, encode_slash = false) {
+    // https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html
+
+    const hex_digit = "0123456789ABCDEF";
+    let result = "";
+
+    for (let i = 0; i < input.length; i++) {
+        const ch : string = input[i];
+
+        if ((ch >= 'A' && ch <= 'Z') || (ch >= 'a' && ch <= 'z') || (ch >= '0' && ch <= '9') || ch == '_' ||
+            ch == '-' || ch == '~' || ch == '.') {
+            result += ch;
+        } else if (ch == '/') {
+            if (encode_slash) {
+                result += "%2F";
+            } else {
+                result += ch;
+            }
+        } else {
+            result += "%";
+            result += hex_digit[ch.charCodeAt(0) >> 4];
+            result += hex_digit[ch.charCodeAt(0) & 15];
+        }
+    }
+    return result;
+}
+
+export function createS3Headers(config : S3Config | undefined, url : string, method : string) : Map<string, string> {
+    const params = getS3Params(config, url, method);
+    return createS3HeadersInternal(params);
+}
+
+export function createS3HeadersInternal(params: S3Params) : Map<string, string> {
+    // this is the sha256 of the empty string, its useful since we have no payload for GET requests
+    const empty_payload_hash = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855";
+
+    const res = new Map<string, string>();
+    // res.set("host", params.host)
+    res.set("x-amz-date", params.datetime_now);
+    res.set("x-amz-content-sha256", empty_payload_hash);
+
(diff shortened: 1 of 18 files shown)


domoritz:
Really cool. How much does this change increase the bundle sizes?

ankoh:
Looks great, thanks a lot!

ankoh:
This PR should not be too severe since we're offloading the requests to the browser xhrs again. But I agree that we should start to track the bundle sizes since we'll otherwise just keep growing.
Also we should think about offering separate bundles with, e.g. optional parquet support.

packages/duckdb-wasm/src/utils/s3_helper.ts
```diff
@@ -0,0 +1,221 @@
...
+
+export function getHTTPUrl(config : S3Config | undefined, url : string) : string {
+    const parsedUrl = parseS3Url(url);
+    const host = parsedUrl.bucket + "." + (config?.endpoint ? config?.endpoint : "s3.amazonaws.com");
+    const http_host = "https://" + host;
+    return http_host + parsedUrl.path;
+}
+
+// this computes the signature from https://czak.pl/2015/09/15/s3-rest-api-with-curl.html
+export function verifyS3Helper() {
```
ankoh, line 165: Can we move this to the test file? Otherwise this will be bundled within the library.

domoritz:
Are you talking abut the WASM or the JS bundles? For the JS bundles, we could look into how well they can be three shaken.  I think WASM doesn't support dynamic modules yet so we have to ship separate bundles unfortunately.

samansmink:
I moved the verify function to the tests and fixed some incorrect casing I found while doing so.

@domoritz Here's the before and after for the /dist, 

file | Master | S3FS branch
-- | -- | --
duckdb-browser-blocking.cjs | 2.2M  | 2.3M 
duckdb-browser-blocking.d.ts | 60B  | 60B 
duckdb-browser-blocking.mjs | 1.6M  | 1.7M 
duckdb-browser-coi.d.ts | 43B  | 43B 
duckdb-browser-coi.pthread.worker.js | 1.0M  | 1.1M 
duckdb-browser-coi.worker.js | 1.1M  | 1.2M 
duckdb-browser-eh.d.ts | 43B  | 43B 
duckdb-browser-eh.worker.js | 1.1M  | 1.2M 
duckdb-browser.cjs | 1.1M  | 1.1M 
duckdb-browser.d.ts | 43B  | 43B 
duckdb-browser.mjs | 833K  | 833K 
duckdb-browser.worker.js | 1.4M  | 1.5M 
duckdb-coi.wasm | 44M  | 44M 
duckdb-eh.wasm | 43M  | 43M 
duckdb-node-blocking.cjs | 1.4M  | 1.4M 
duckdb-node-blocking.d.ts | 57B  | 57B 
duckdb-node-coi.d.ts | 43B  | 43B 
duckdb-node-eh.d.ts | 43B  | 43B 
duckdb-node-eh.worker.cjs | 1.4M  | 1.4M 
duckdb-node.cjs | 157K  | 157K 
duckdb-node.d.ts | 43B  | 43B 
duckdb-node.worker.cjs | 1.1M  | 1.1M 
duckdb.wasm | 46M  | 46M 
tests-browser.js | 2.5M  | 2.7M 
tests-node.cjs | 2.1M  | 2.1M 
types | 128B  | 128B

ankoh:
2.2MB? That's the debug build with inlined sourcemap no?

samansmink:
@ankoh ah yes it is indeed

domoritz:
I also found that esbuild isn't very good at treeshaking/ dead code elimination so you could try running terser to see whether that makes our bundles smaller.

ankoh:
True, we could revisit the js bundling as well in a concentrated effort to get our package sizes down. That's not a priority for me in the next weeks but we should do that.

Esbuild is just so fast, though. :) But introducing terser just for the release builds would be an option.

gordonje:
Hello, I came here looking for exactly this feature (I think). I'm hoping to have the ability to read parquet files from a private AWS S3 bucket in Observable notebooks. 

I expect the cell declaring a client for a loaded duckDB would work something like this:

```js
client = {
  const client = new DuckDBClient();

  await client.query(`
    SET s3_access_key_id='${Secret("AWS-Access-Key")}';
    SET s3_secret_access_key='${Secret("AWS-Secret-Key")}';
    SET s3_region='us-east-1';

    CREATE TABLE my_table AS 
      SELECT * 
      FROM parquet_scan([
      's3:/my-bucket/path/to/my/results/0000_part_00.parquet',
      's3:/my-bucket/path/to/my/results/0000_part_00.parquet'
    ]);
`);
  
  return client;
}
```

Will follow your repo in anticipation of the next release.

Thank so much for sharing your excellent work with the world.

ankoh:
@gordonje you can try it out in **1.13.1-dev67.0** but you'll have to send the SET commands with separate `query` calls for now.

gordonje:
It worked! 

Specifically something that looked like this:

```js
dbClient = {
  const client = new DuckDBClient();

  await client.query(
    `SET s3_access_key_id='${Secret("AWS-Access-Key")}';`
  );
  await client.query(
    `SET s3_secret_access_key='${Secret("AWS-Secret-Key")}';`
  );
  await client.query(
    `SET s3_region='us-east-1';`
  );

  await client.query(`
    CREATE TABLE my_table AS 
      SELECT * 
      FROM parquet_scan([
      's3://path/to/my/results/0000_part_00.parquet',
      's3://path/to/my/results/0001_part_00.parquet',
      's3://path/to/my/results/0002_part_00.parquet'
    ]);
`);

  return client;
}
```

However, when I encountered `Error: 6657936` when I tried to use a glob pattern:

```js
dbClient = {
  const client = new DuckDBClient();

  await client.query(
    `SET s3_access_key_id='${Secret("AWS-Access-Key")}';`
  );
  await client.query(
    `SET s3_secret_access_key='${Secret("AWS-Secret-Key")}';`
  );
  await client.query(
    `SET s3_region='us-east-1';`
  );

  await client.query(`
    CREATE TABLE addresses AS 
      SELECT * 
      FROM parquet_scan('s3://path/to/my/results/*.parquet');
  `);

  return client;
}
```

ankoh:
Glad you got that working!
And we don't support glob for s3 yet.

gordonje:
Got it. Regardless, I'm very excited about this enhancement, and I expect to be making more use of duckDB in future work. Thank you again.

kimmolinna:
Is this working also for example Scaleway which is based on the S3 protocol?

PR #562, opened 2022-01-21
S3 writes & local s3 test server for CI

samansmink:
Implemented basic implementation for writing  HTTP/S3 files. The files are converted to buffers on opening to allow writes. On closing the file we upload the buffer synchronously with a single HTTP PUT request. For large uploads we should look into multipart uploads. I can look into that after implementing s3 writes in DuckDB as well.

I also added a local test s3 server using [S3rver](https://www.npmjs.com/package/s3rver). It seems to work quite well, but as discussed with @ankoh, we should keep an eye on its stability and possibly change it for something more solid such as [minio]{https://min.io/}. switching to minio should be quite simple but complicate the local dev

Note that s3rver only really seems to validate the credentials and does not actually check the authorization signature, therefore I added a testcase to the manual validation of the auth header to make this a bit more solid.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@ compile_commands.json
 .vscode
 !.vscode/settings.json
 data/tpch/**/*.parquet
+.idea
+lib/.idea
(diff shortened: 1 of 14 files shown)


ankoh (approved):

ankoh:
Awesome, thx!

david542542:
> For large uploads we should look into multipart uploads. I can look into that after implementing s3 writes in DuckDB as well.

Hi @samansmink, @ankoh -- not sure if this is helpful, but we developed and used this a ways back to upload large video files (1TB) to s3 using threads with webworkers: https://github.com/cinely/mule-uploader. Though I'm sure there are lots of similar things that you could use but see if that might work for this.

samansmink:
Hey @david542542 thanks for the suggestion! That definitely looks interesting for us!

PR #594, opened 2022-02-04
fix: s3 signing bug

samansmink:
There was a bug with the signing of S3 requests to endpoints other than the default s3.amazonaws.com as the host for signing would not be set correctly. Should be fixed now.


ankoh:
Awesome, thx!

PR #606, opened 2022-02-09
UDF extensions part one

hannesmuehleisen:
- UDFs can now have 0, 1, 2, or 3 arguments
- Arguments can be `NULL`, return values can be `NULL`
- `VARCHAR` and `DOUBLE` arguments and return values, also `TINYINT`/`SMALLINT`/`INTEGER` and their unsigned versions

### DIFF
--- a/packages/duckdb-wasm/src/bindings/udf_runtime.ts
+++ b/packages/duckdb-wasm/src/bindings/udf_runtime.ts
@@ -97,7 +97,7 @@ export function callScalarUDF(
         // create argument arrays
         for (const idx in desc.args) {
             const arg = desc.args[idx];
-            var arg_arr = ptr_to_arr(mod, Number(ptrs[arg.data_buffer]), arg.physical_type, desc.rows);
+            const arg_arr = ptr_to_arr(mod, Number(ptrs[arg.data_buffer]), arg.physical_type, desc.rows);
             const validity = ptr_to_arr(mod, Number(ptrs[arg.validity_buffer]), 'UINT8', desc.rows);
 
             if (arg_arr.length == 0 || validity.length == 0) {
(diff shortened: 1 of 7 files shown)


packages/duckdb-wasm/src/bindings/udf_runtime.ts
```diff
@@ -14,36 +14,212 @@ function storeError(mod: DuckDBModule, response: number, message: string) {
...
+        }
+
+        // schema description as json
+        const desc_str = new TextDecoder().decode(mod.HEAPU8.subarray(descPtr, descPtr + descSize));
+        const desc = JSON.parse(desc_str);
+        //console.log(desc);
+
+        // array of buffer pointers referred to from schema
+        const ptrs_buf = mod.HEAPU8.subarray(ptrsPtr, ptrsPtr + ptrsSize);
+        const ptrs = new BigUint64Array(ptrs_buf.buffer, ptrs_buf.byteOffset, ptrsSize / 8);
```
ankoh, line 93: We might want to change this one later.
https://caniuse.com/?search=BigUint64Array

We do a platform check for BigIntArrays in the remaining API.
https://github.com/duckdb/duckdb-wasm/blob/d050191e1cad0525b40d466d714610543371739f/packages/duckdb-wasm/src/platform.ts#L67

ankoh:
Awesome, thx 👍

ankoh (approved):

packages/duckdb-wasm/src/bindings/udf_runtime.ts
hannesmuehleisen: Ah missed that one, will just switch them to Double, too, seems to work fine

Y--:
Awesome! Great work :-)

PR #614, opened 2022-02-11
Enable FTS extension

hannesmuehleisen:
This adds experimental support for the FTS extension in WASM.

@ankoh is there a way we can make this available, would it mean more binary blobs?


Y-- (approved):
Thanks :-)

ankoh (approved):

PR #900, opened 2022-07-21
refactor: use tableToIPC directly

domoritz:
So we are consistent with the async bindings.


packages/duckdb-wasm/src/bindings/connection.ts
```diff
@@ -78,9 +78,7 @@ export class DuckDBConnection {
 
     /** Insert an arrow table */
     public insertArrowTable(table: arrow.Table, options: ArrowInsertOptions): void {
-        const writer = new arrow.RecordBatchStreamWriter();
-        writer.writeAll(table);
-        const buffer = writer.toUint8Array(true);
+        const buffer = arrow.tableToIPC(table);
```
domoritz, line 81:
```suggestion
	const buffer = arrow.tableToIPC(table, 'stream');
```

PR #1009, opened 2022-09-18
Use MANUAL_BUNDLES where applicable

alexisweill:
Replace JSDELIVERY_BUNDLES by MANUAL_BUNDLES where applicable
I was not able to test the changes but reading the code it seems like in the last 3 examples it should be referring to MANUAL_BUNDLES

### DIFF
--- a/packages/duckdb-wasm/README.md
+++ b/packages/duckdb-wasm/README.md
@@ -52,7 +52,7 @@ const MANUAL_BUNDLES: duckdb.DuckDBBundles = {
     },
 };
 // Select a bundle based on browser checks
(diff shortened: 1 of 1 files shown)


ankoh:
Thanks for the PR!
But this in intentional.
I'd prefer that people stick to the Jsdelivr bundles by default.
The chances that those will work out of the box are much higher as the bundler does not need to take care of the wasm modules.
Some bundlers out there do not really treat wasm was raw binary assets which will complicate things a lot (and has already raised quite a few questions in the discord).

Also, loading from jsdelivr shares the wasm modules among sites which is also conceptually quite nice.

PR #1025, opened 2022-10-05
fix typo in docs, exclamation is invalid

llimllib:


llimllib:
oh - the documentation assumes you're using Typescript!

I'll close this, but the doucmentation should probably mention that because anybody who tries to copy the code into a javascript file is likely to be confused

PR #1075, opened 2022-11-27
feat: access mode config

Mause:
- feat: add accessMode config param
- chore: fix lint issue

### DIFF
--- a/lib/include/duckdb/web/config.h
+++ b/lib/include/duckdb/web/config.h
@@ -2,6 +2,7 @@
 #define INCLUDE_DUCKDB_WEB_CONFIG_H_
 
 #include <rapidjson/document.h>
+#include "duckdb/main/config.hpp"
 
 #include <cstdint>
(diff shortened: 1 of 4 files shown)


Mause:
It looks like the build is currently failing as the version of arrow we're using doesn't include https://github.com/apache/arrow/commit/1a00075115ed3487e36281f29c6a7911a7d2bf8d, but updating to one that does causes other compilation issues

ankoh:
The ci script incorrectly shows a green build step and then fails in the next one.
That's the error:
```
../../src/config.cc:105:33: error: assigning to 'uint8_t' (aka 'unsigned char') from incompatible type 'duckdb::AccessMode'
            config.accessMode = ParseAccessMode(doc["access_mode"].GetString());
```

Mause:
Closing in favour of https://github.com/duckdb/duckdb-wasm/pull/1260

PR #1120, opened 2023-01-11
Add origin-trial credentialless credentials

carlopi:


domoritz:
What's this origin trial for?

carlopi:
According to https://developer.chrome.com/blog/enabling-shared-array-buffer/#origin-trial, it should allow back SharedArrayBuffer iff:
* you are on a given domain (and someone has registered to get the token + advertise it either via <meta> tag or HTTP header)
* user is browsing from Desktop + Chrome-based browser + stars align (as in, Chrome reserves the right not to enable it for A/B testing or whatever reason).
I haven't run any real experiment with this, might be way off, tomorrow will take a proper look playing with this on a domain I have control on and then come back.

This is very much a work-around / experiment, but might be worth to have threading enabled even just on the shell for demo purposes.

PR #1151, opened 2023-02-15
Update to duckdb 070

carlopi:
Huge thanks to @kimmolinna for commits 1-5, bumping duckdb version to 0.7.0 and fixing consequent API breakages.

There is something going on with testing infrastructure that I have to come back on, failures were all like:
`Expected 1 to equal 1.` (??) potentially it has to do with the testing infrastructure itself, I am checking this out to be able to revert the last commit.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -329,7 +329,7 @@ endif()
 
 if(NOT EMSCRIPTEN)
   set(TEST_CC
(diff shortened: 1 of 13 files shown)


Mytherin (approved):
Thanks! The C++ changes look good to me - it would be great if we could get the tests uncommented out though as you mentioned.

kimmolinna:
And there should be tests also for different storage provider connections.

PR #1174, opened 2023-03-03
Upstream a few changes we made internally

Y--:
(see comments inline)

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -19,7 +19,7 @@ endif()
 # ---------------------------------------------------------------------------
 # Ccache
 
-if(EMSCRIPTEN)
+if(EMSCRIPTEN AND EXISTS "$ENV{EMSDK}/ccache/git-emscripten_64bit/bin")
   find_program(CCACHE_PROGRAM ccache
                HINTS $ENV{EMSDK}/ccache/git-emscripten_64bit/bin)
 else()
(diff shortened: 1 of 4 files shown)


lib/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ endif()
 # ---------------------------------------------------------------------------
 # Ccache
 
-if(EMSCRIPTEN)
+if(EMSCRIPTEN AND EXISTS "$ENV{EMSDK}/ccache/git-emscripten_64bit/bin")
```
Y--, line 22: This allows compilation on ARM64 plaforms

packages/duckdb-wasm/bundle.mjs
```diff
@@ -175,7 +175,7 @@ fs.copyFile(path.resolve(src, 'bindings', 'duckdb-coi.wasm'), path.resolve(dist,
         bundle: true,
         minify: true,
         sourcemap: is_debug ? 'inline' : true,
-        external: EXTERNALS_WEBWORKER,
+        external: [ EXTERNALS_WEBWORKER, 'module' ],
```
Y--, line 178: This is a an initial fix for supporting recent emscripten

lib/include/duckdb/web/arrow_stream_buffer.h
```diff
@@ -85,3 +88,5 @@ struct BufferingArrowIPCStreamDecoder : public arrow::ipc::StreamDecoder {
 
 }  // namespace web
 }  // namespace duckdb
+
+#endif // INCLUDE_DUCKDB_WEB_ARROW_STREAM_BUFFER_H_
```
ankoh, line 92: Just need a space here for clang-format

ankoh (changes requested):

lib/include/duckdb/web/arrow_stream_buffer.h
Y--: Fixed, thanks :-)

PR #1194, opened 2023-03-17
Fixes

carlopi:
Commits are unconnected.

* shell.ts -> I saw in some local configurations numbers out of 0-10
* npm's README was off, pointing to a specific version instead of latest
* GitHub and AWS returns 200 when range requests are valid, adapting to avoid full requests (note that accept-bytes is not accessible in most servers, so can't be checked

### DIFF
--- a/packages/duckdb-wasm-shell/src/shell.ts
+++ b/packages/duckdb-wasm-shell/src/shell.ts
@@ -90,7 +90,7 @@ export async function embed(props: ShellProps) {
     // Progress handler
     const progressHandler = (progress: InstantiationProgress) => {
         if (progress.bytesTotal > 0) {
-            const blocks = Math.floor((progress.bytesLoaded / progress.bytesTotal) * 10.0);
+            const blocks = Math.max(Math.min(Math.floor((progress.bytesLoaded / progress.bytesTotal) * 10.0), 10.0), 0.0);
             const bar = `${'#'.repeat(blocks)}${'-'.repeat(10 - blocks)}`;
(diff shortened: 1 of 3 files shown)


packages/duckdb-wasm-shell/src/shell.ts
```diff
@@ -90,7 +90,11 @@ export async function embed(props: ShellProps) {
     // Progress handler
     const progressHandler = (progress: InstantiationProgress) => {
         if (progress.bytesTotal > 0) {
-            const blocks = Math.floor((progress.bytesLoaded / progress.bytesTotal) * 10.0);
+            var blocks = Math.floor((progress.bytesLoaded / progress.bytesTotal) * 10.0);
```
ankoh, line 93: Please use let instead of var

packages/duckdb-wasm-shell/src/shell.ts
```diff
@@ -90,7 +90,11 @@ export async function embed(props: ShellProps) {
...
+            if (blocks < 0.0)
```
ankoh, line 94: Brackets or max(blocks, 0.0) please

ankoh (changes requested):

carlopi:
Updated the PR with the comments + adding a fall back for detecting whether range requests are accepted: asking explicitly for range 0-0 (instead of 0-) will give back 206 from raw.githubusercontent.com (and potentially other servers) at the cost of not providing *** full request length.

Proper solution for this would be a rework of checkFile, so that checkFile will also save/cache the length of the resource, but this is for a later refactor.

*** = Full request length is there, but filtered out by the browser

ankoh (approved):

PR #1198, opened 2023-03-20
Fixes: actually fix #1189 + bump duckdb master

### LINKED ISSUE #1189: Restrictive logic for checking whether a remote parquet file supports range queries
## Issue
When I attempt to query a parquet file I have hosted myself with the DuckDB WASM web shell (https://shell.duckdb.org/), it seems to always fallback to a full HTTP read.

## Cause
I have done a bit of testing and this appears to be because when DuckDB sends its initial `HEAD` request with the Range header, it is expecting an HTTP Response Code of `206`:
https://github.com/duckdb/duckdb-wasm/blob/cceb0ed3db100cebdd5eeecc3178baf3ba912313/packages/duckdb-wasm/src/bindings/runtime_browser.ts#L168

### What works
With the sample file that is provided in the docs (https://shell.duckdb.org/data/tpch/0_01/parquet/customer.parquet), a `HEAD` request with a Range header does return a status code of `206`. However, this response does not seem to be a universal standard.

### What doesn't work
I have tested with two hosted files, one on Github Pages ([here](https://raw.githubusercontent.com/carstonhernke/demo-h3-data/main/parquet/germany_population.parquet)), and one served with AWS Cloudfront ([here](https://honeycomb.carston.org/data/usa_population.parquet)). They both seem to return a status code of `200` for a `HEAD` request with a Range header. **This causes the check above to be false and triggers the fallback to a full HTTP read**. However, both these hosting providers do support HTTP Range requests.

## Potential solutions
I can think of a few solutions here:
- Enhance the logic to test whether a given file can be accessed with Range requests (not an HTTP status code expert, but if the `Accept-Ranges` header is present on the response with a value of `bytes`, this should be enough, right?)
- Add some better documentation explaining the `HEAD` request process and necessary headers, so that users can configure their server/ hosting provider settings accordingly.

carlopi:
I though I had tested this, but a GET request is actually needed in place of a HEAD request to trigger the ranges.

### DIFF
--- a/packages/duckdb-wasm/src/bindings/runtime_browser.ts
+++ b/packages/duckdb-wasm/src/bindings/runtime_browser.ts
@@ -171,10 +171,10 @@ export const BROWSER_RUNTIME: DuckDBRuntime & {
                                 // Send a dummy range request querying the first byte of the file
                                 const xhr2 = new XMLHttpRequest();
                                 if (file.dataProtocol == DuckDBDataProtocol.S3) {
-                                    xhr2.open('HEAD', getHTTPUrl(file.s3Config, file.dataUrl!), false);
-                                    addS3Headers(xhr2, file.s3Config, file.dataUrl!, 'HEAD');
+                                    xhr2.open('GET', getHTTPUrl(file.s3Config, file.dataUrl!), false);
+                                    addS3Headers(xhr2, file.s3Config, file.dataUrl!, 'GET');
                                 } else {
-                                    xhr2.open('HEAD', file.dataUrl!, false);
+                                    xhr2.open('GET', file.dataUrl!, false);
                                 }
                                 xhr2.setRequestHeader('Range', `bytes=0-0`);
                                 xhr2.send(null);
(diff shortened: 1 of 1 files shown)


carlopi:
I actually ended up reworking the openFile. Documenting the logic:

```
1. HEAD with header range 0-.
     -> 206 + contentLength? Range supported, return
2. IFF full download are not supported, bail out.
3. GET with header range 0-0
     -> 206 + contentLength=1? Range supported, return
     -> 200 + got the full file already? Copy response, return
4. GET with no headers
     -> 200, Copy response, return
5. Failure
```

It seems to make sense to me. Probably we should also bail out earlier on non 20x status (although we should be guarded by checkFile, and we should consider caching during checkFile the 206 + contentRange (that will avoid the HEAD request).

Current situation is 2 HEAD for cases like shell.duckdb.com, 2 HEAD + 1 GET for the raw.githubusercontent.com case, and 2 HEAD + 1 GET in the failure cases.
Caching will cut 1 HEAD request on every case, that mostly translates to minor improvement on latency. While situation pre-last commit was 2 HEAD, 2 HEAD + 1 GET, 2 HEAD + 2 GET.

Issue #1014, opened 2022-09-24
Shell is stuck loading when using Firefox Incognito mode


ttomasz:
When opening https://shell.duckdb.org/ in Firefox Incognito mode the shell gets stuck like in this screenshot:
![image](https://user-images.githubusercontent.com/31594210/192120858-d52ec191-9d82-470b-bdca-e95cade7c2df.png)
In regular Firefox window it works fine:
![image](https://user-images.githubusercontent.com/31594210/192120889-22e6ad54-dc90-4c58-a2eb-26490a38e290.png)

Linux, Firefox 105.0.1

carlopi:
Seems it's actually fixed, thanks again for opening this.

Behaviour now on Firefox incognito, and to some degree situations where indexedDB will fail to initialise, is that a console.warn("") message will inform no history could be loaded/persisted, but the shell will work regularly.

Thanks @ankoh and @domoritz.

PR #1208, opened 2023-03-23
Remove dependency on indexedDB

carlopi:
This allow duckdb-wasm-shell to work also on Firefox Incognito, fixing #1014

### DIFF
--- a/packages/duckdb-wasm-shell/src/utils/history_store.ts
+++ b/packages/duckdb-wasm-shell/src/utils/history_store.ts
@@ -21,12 +21,14 @@ export class HistoryStore {
     protected _idb: IDBDatabase | null;
     protected _nextEntryKey: number;
     protected _entryCount: number;
+    protected _isWorking: boolean;
 
     public constructor() {
         this._idbFactory = window.indexedDB;
         this._idb = null;
         this._nextEntryKey = 0;
         this._entryCount = 0;
+        this._isWorking = (this._idbFactory != null);
     }
 
     /// Load entire history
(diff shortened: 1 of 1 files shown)


packages/duckdb-wasm-shell/src/utils/history_store.ts
```diff
@@ -21,12 +21,14 @@ export class HistoryStore {
...
     protected _nextEntryKey: number;
     protected _entryCount: number;
+    protected _isWorking: boolean;
 
     public constructor() {
         this._idbFactory = window.indexedDB;
         this._idb = null;
         this._nextEntryKey = 0;
         this._entryCount = 0;
+        this._isWorking = (this._idbFactory != null);
```
domoritz, line 31: Instead of this state variable, would it make sense to return a different instance that has a different implementation of the methods that are not supported?

carlopi: Thanks for the review!
After fixing it, I think actually the minimal set of changes we could get away with for fixing Firefox Incognito behaviour are:
```
            req.onerror = err => {
                console.log("Error while opeing indexedDB connection", err);
                // NOTE: we are using resolve(null) on purpose, since null is a vaild value,
		//	and reject (un-catched) will trigger a failure
                resolve(null);
            };
```
instead of
```
req.onerror = err => reject(err);
```
+ doing the checks on `loadMetadata`, `reset` and `push` on _idb being initialzied (instead of adding a variable).

carlopi: Basically applying this commit: https://github.com/duckdb/duckdb-wasm/compare/master...carlopi:duckdb-wasm:fixIncognito2 instead of the current PR.

domoritz: I don't think I like either as they are both swallowing errors or intransparently change behavior. 

Can we just wrap the call where we initialize the history db to either return an instance of this class or an empty mock?

carlopi:
This is not the right solution, moving this to #1209

PR #1209, opened 2023-03-23
Be flexible of initialization problems in indexedDB

carlopi:
This allow duckdb-wasm-shell to work also on Firefox Incognito, fixing #1014

This is an alternative/better implementation of #1208

### DIFF
--- a/packages/duckdb-wasm-shell/src/utils/history_store.ts
+++ b/packages/duckdb-wasm-shell/src/utils/history_store.ts
@@ -71,7 +71,12 @@ export class HistoryStore {
                 const idb = req.result;
                 resolve(idb);
             };
-            req.onerror = err => reject(err);
+            req.onerror = err => {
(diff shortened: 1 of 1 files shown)


ankoh (approved):

packages/duckdb-wasm-shell/src/utils/history_store.ts
```diff
@@ -71,14 +71,21 @@ export class HistoryStore {
                 const idb = req.result;
                 resolve(idb);
             };
-            req.onerror = err => reject(err);
+            req.onerror = err => {
+                console.log("Error while opeing indexedDB connection", err);
+                // NOTE: we are using resolve(null) on purpose, since null is a vaild value,
+		//	and reject (uncatched) will trigger a failure
```
domoritz, line 77: The indentation here is different btw.

Issue #1218, opened 2023-03-29
"make" fails


whscullin:
After https://github.com/duckdb/duckdb-wasm/pull/1211, running "make" now gives me:

```
make: *** No rule to make target 'docker_ci_image', needed by 'wrapped_wasm_caches'.  Stop.
```

The `docker_ci_image` was changed to `build/docker_ci_image`, but most places still reference `docker_ci_image`. This is using GNU Make 4.3.

PR #1224, opened 2023-03-31
Makefile rework /2

carlopi:
As noticed by [whscullin](https://github.com/whscullin) in https://github.com/duckdb/duckdb-wasm/issues/1218, running `make` as been non-working since #1211.

This PR fixes that plus introduces a CI test on basic makefile functionality (for now "make" and "make clean && make").

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,7 +149,7 @@ lib_debug: lib
 
 bench_build:
 	yarn workspace @duckdb/benchmarks build
-	touch bench_build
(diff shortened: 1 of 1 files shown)


ankoh:
I think the mentioned CI test is missing?

carlopi:
Yes, still missing, I am failing on some missed packages, would need another couple of iterations, but this can be merged as in works locally (while currently it's somehow broken).

PR #1227, opened 2023-04-03
Safer uniq ptr

carlopi:
Demo of using https://github.com/duckdb/duckdb/pull/6950

(IF that happens to be merged, last commit can be dropped and this can be finalized)

### DIFF
--- a/lib/src/json_table.cc
+++ b/lib/src/json_table.cc
@@ -359,7 +359,7 @@ std::unique_ptr<duckdb::ArrowArrayStreamWrapper> TableReader::CreateStream(uintp
     auto reader_copy = (*reader)->CloneShared();
 
     // Create arrow stream
-    auto stream_wrapper = duckdb::make_unique<duckdb::ArrowArrayStreamWrapper>();
+    auto stream_wrapper = std::make_unique<duckdb::ArrowArrayStreamWrapper>();
     stream_wrapper->arrow_array_stream.release = nullptr;
     auto maybe_ok = arrow::ExportRecordBatchReader(reader_copy, &stream_wrapper->arrow_array_stream);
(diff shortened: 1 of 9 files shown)


lib/src/json_table.cc
```diff
@@ -379,7 +379,7 @@ void TableReader::GetSchema(uintptr_t this_ptr, duckdb::ArrowSchemaWrapper& sche
     auto reader_copy = (*reader)->CloneShared();
 
     // Create arrow stream
-    auto stream_wrapper = duckdb::make_unique<duckdb::ArrowArrayStreamWrapper>();
+    auto stream_wrapper = std::make_unique<duckdb::ArrowArrayStreamWrapper>();
```
Tishj, line 382: I'm confused, this was changed from `duckdb::make_unique` to `std::make_unique`
instead of using `duckdb::make_uniq` ?

lib/src/json_table.cc
```diff
@@ -359,7 +359,7 @@ std::unique_ptr<duckdb::ArrowArrayStreamWrapper> TableReader::CreateStream(uintp
...
```
Tishj, line 362: Same as previous comment

Tishj:
I completely understand what this PR is doing now, and it's great 👍 

To summarize:
`safe_unique_ptr` will deprecate (remove) `duckdb::make_unique`, by removing the use of `duckdb::make_unique` we can make incremental changes that make sure there is a small amount of steps we can take to go from one working state to the next

carlopi:
Thanks @Tishj!

I put here a basic diagram / explanation : https://github.com/duckdb/duckdb/pull/6950#issuecomment-1493948484 for whoever passes next over this.

Will update to actual PR when ready.

PR #1232, opened 2023-04-05
Load loadable

carlopi:
More work to support both loadable and non-loadable extensions, plus updating duckdb to latest


Mytherin (approved):
Thanks for the PR! LGTM

Issue #1235, opened 2023-04-06
Wasm range request failure


taniabogatsch:
### What happens?

duckdb-wasm fails when requesting data from the GitHub API during range requests.

### To Reproduce

Go to https://shellwip.duckdb.org.
```sql
LOAD json;
SELECT * FROM read_json_auto('https://api.github.com/repos/duckdb/duckdb/commits?per_page=10000&foo.json');
```

```
Invalid Error: Error: Range request for https://api.github.com/repos/duckdb/duckdb/commits?per_page=10000&foo.json failed with error: Error: Range request for https://api.github.com/repos/duckdb/duckdb/commits?per_page=10000&foo.json did not return a partial response: 200 """
```

### Browser/Environment:

Safari Version 16.4

### Device:

MacBook Pro M1 2020

### DuckDB-Wasm Version:

1.11.0

### DuckDB-Wasm Deployment:

shellwip.duckdb.org

### Full Name:

Tania Bogatsch

### Affiliation:

DuckDB Labs

carlopi:
This has been now fixed via #1236, builds are in progress.

Thanks!

PR #1236, opened 2023-04-06
HTTP: accept 200 in place of 206, truncating content

carlopi:
This solves https://github.com/duckdb/duckdb-wasm/issues/1235.

Keeping track in an internal cache of the additional bytes would be beneficial, TODO added.
In any case this piggy-backing on browser-cache is already be a step further than current situation.

### DIFF
--- a/packages/duckdb-wasm/src/bindings/runtime_browser.ts
+++ b/packages/duckdb-wasm/src/bindings/runtime_browser.ts
@@ -395,9 +395,12 @@ export const BROWSER_RUNTIME: DuckDBRuntime & {
                             mod.HEAPU8.set(src, buf);
                             return src.byteLength;
                         } else if (xhr.status == 200) {
-                            throw new Error(
-                                `Range request for ${file.dataUrl} did not return a partial response: ${xhr.status} "${xhr.statusText}"`,
-                            );
+                            // TODO: here we are actually throwing away all non-relevant bytes, but this is still better than failing
+                            //       proper solution would require notifying duckdb-wasm cache, while we are piggybackign on browser cache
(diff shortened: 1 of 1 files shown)


packages/duckdb-wasm/src/bindings/runtime_browser.ts
```diff
@@ -395,9 +395,11 @@ export const BROWSER_RUNTIME: DuckDBRuntime & {
                             mod.HEAPU8.set(src, buf);
                             return src.byteLength;
                         } else if (xhr.status == 200) {
-                            throw new Error(
-                                `Range request for ${file.dataUrl} did not return a partial response: ${xhr.status} "${xhr.statusText}"`,
-                            );
+                            // TODO: here we are actually trowing away all non-relevant bytes, but this is still better than failing
```
samansmink, line 398: nit: typo in `trowing`

as discussed on discord already, we probably want a warning for the user here

carlopi:
Expanding for posterity:

There are cases where ducdkb-wasm HTTP backend decides a range request is expected to work, and perform that request.

Detection might be unreliable, so server might send back full content (200) instead of the requested range (206).

Current handling: throw an error (that would bubble to make the read fail).

Proposed change: satisfy the read, at the cost of throwing away useful bytes, with a console.warn.
Problematic case is multiple small reads that all imply a full file download. Two things that alleviate this are presence of browser cache in the common browser setup AND the fact that read size keeps increasing so in practice only single digit (log (N / constant) reads might be actually needed).

A proper solution, that would requires a rework of the HTTP backend, would be notifying the rest of the system that those bytes are now known. But I believe for now the incremental change it's the better solution, and this will need to be taken a look again while unifying behaviour between duckdb and duckdb-wasm HTTP backends.

Mytherin (approved):

Mytherin:
Thanks!

PR #1237, opened 2023-04-06
Safe vector

Tishj:
This PR should make the minimum changes required to work with the safe_vector PR in duckdb https://github.com/duckdb/duckdb/pull/6927


carlopi (approved):
Thanks a lot for filing this!

Context: DUCKDB_API returns duckdb::vector, that happens to be an alias to std::vector, but that it's going to change and this PR makes duckdb-wasm compatible both with current duckdb AND with future changes to be soon merged.

ankoh (approved):

PR #1245, opened 2023-04-11
Bump to GET ranged request if HEAD fails

carlopi:
This would need to be reviewed together with the broader partial unification with HTTPFS, since many codepaths should be abstracted in the same functions, behaviour should be more consistent.

Worst case here is a server that don't do neither HEAD nor ranged requests, so send back the full content, and the user was only interested in some aggregate statistic.
But I would argue that requesting the resources implies some kind of intention, that couple with more visible warnings (still not there) should make it bearable.

### DIFF
--- a/packages/duckdb-wasm/src/bindings/runtime_browser.ts
+++ b/packages/duckdb-wasm/src/bindings/runtime_browser.ts
@@ -286,8 +286,31 @@ export const BROWSER_RUNTIME: DuckDBRuntime & {
                 }
                 xhr.send(null);
                 if (xhr.status != 200 && xhr.status !== 206) {
-                    failWith(mod, `HEAD request failed: ${path}`);
-                    return;
+                    // Pre-signed resources on S3 in common configurations fail on any HEAD request
+                    // https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/s3-example-presigned-urls.html
+                    // so we need (if enabled) to bump to a ranged GET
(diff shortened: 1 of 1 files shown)


samansmink (commented):
I'm ok with this, but would prefer to see `allowFullHttpReads` as default off. We could consider adding this as a SET variable in duckdb similiar to how we did the s3 credentials: https://github.com/duckdb/duckdb-wasm/blob/e0bf0f326fe0f8db999e00f2f5b9c98a2e5ae657/lib/src/webdb.cc#L668 Then we could have the error message tell people to turn this on at their own risk, so they will know where to look when then realize things are slow.

samansmink:
@Alex-Monahan That's a fair point, but I think the differences here could be very big, as in multiple orders of magnitude, completely prohibitively big. We're potentially downloading the whole file over and over many times. This could cause users to either post a bug report or abandon duckdb for their workload. We could just have the error message say exactly what command to enter to enable the full download. The thing is, in many cases the user shouldnt actually enable it, but ensure the endpoint supports range requests.

carlopi:
I think in this particular case we are only trying to check whether a file exist, so it should be performed only once per file (per query???), and I would assume the most likely case here is still getting back the range request as expected.

Many things would be improvements (async I/O, via proper support or threads + dedicated worker, or ...), and more control to end user is the proper way out, but I feel for now yet another patch might be needed here.

Having default being tunable by users via SQL seems a great point in any case, will look into implementing that, and after that plus more visible warnings we could turn the setting on (potential) full range requests off by default.

PR #1252, opened 2023-04-19
Fix #1251: readFile should be robust to empty ranges

### LINKED ISSUE #1251: Invalid Range request - 460-459 , registerFileUrl / IMPORT DATABASE
### What happens?

Loading a CSV via registerfileURl and then IMPORT DATABASE results in an invalid range request.

console message:
FAIL WITH: Error: Range request for http://10.0.12.221:8080/db/raw_csv__test.csv failed with error: Error: Range request for http://10.0.12.221:8080/db/raw_csv__test.csv did returned non-success status: 416 "Range Not Satisfiable""

chrome DevTools reports the request as Range: bytes=460-459

webserver is nginx (on windows) stable: nginx/1.24.0

### To Reproduce

register files from an EXPORT DATABASE './db'; command

```
    await db.registerFileURL(`./import/schema.sql`, 'http://10.0.12.221:8080/db/schema.sql',duckdb.DuckDBDataProtocol.HTTP, true);
    await db.registerFileURL(`./import/load.sql`, 'http://10.0.12.221:8080/db/load.sql',duckdb.DuckDBDataProtocol.HTTP, true);
    await db.registerFileURL(`./import/raw_csv__test.csv`, 'http://10.0.12.221:8080/db/raw_csv__test.csv',duckdb.DuckDBDataProtocol.HTTP, true);
```

try IMPORT DATABASE './public' ; 
```
 const c = await db.connect(); 
await c.query(`IMPORT DATABASE './import' `)
```

schema.sql and load.sql are fetched correctly.
the csv file however is fetched 
first with a HEAD
``` Range: bytes=0- ``` 
response: 
```
Content-Length: 460
Content-Range: bytes 0-459/460
Content-Type: text/csv
```
2nd request GET
``` Range: bytes=0-459 ```
response (the entire file)
``` 
Content-Length: 460
Content-Range: bytes 0-459/460
Content-Type: text/csv
```
but then a 3rd request is issued: GET
```Range: bytes=460-459```
which fails with a ```416 Requested Range Not Satisfiable```

### Browser/Environment:

Chrome Version 112.0.5615.121 (Official Build) (64-bit)

### Device:

windows 10 pc

### DuckDB-Wasm Version:

'v0.7.2-dev1385', 'e8b7132504'

### DuckDB-Wasm Deployment:

https://cdn.jsdelivr.net/npm/@duckdb/duckdb-wasm/+esm

### Full Name:

Jo Degraef

### Affiliation:

none

carlopi:
In case where bytes is 0, read will potentially issue a ranged request on (location, location-1), and this will end up in a failure.
Returning 0 means that 0 bytes have been read, so it's sort of consistent.

I haven't got a local repro (yet), then there is to investigate also why readFile on an empty range is issued.

### DIFF
--- a/packages/duckdb-wasm/src/bindings/runtime_browser.ts
+++ b/packages/duckdb-wasm/src/bindings/runtime_browser.ts
@@ -389,6 +389,10 @@ export const BROWSER_RUNTIME: DuckDBRuntime & {
         return 0;
     },
     readFile(mod: DuckDBModule, fileId: number, buf: number, bytes: number, location: number) {
+        if (bytes == 0) {
+            // Be robust to empty reads
+            return 0;
+        }
         try {
             const file = BROWSER_RUNTIME.getFileInfo(mod, fileId);
             switch (file?.dataProtocol) {


Mytherin (approved):

Mytherin:
Thanks! LGTM

PR #1261, opened 2023-04-28
Add Javascript array example

bitflower:


bitflower:
As discussed on Discord: Will test this first, then come back to see if the code is 100% correct.

domoritz (changes requested):
I don't think this is the best approach. It's more efficient to instead create an arrow table from the array and then inserting the arrow table.

bitflower:
Are you referring to `apache-arrow` @domoritz ? Because this would introduce another dependency just for reading a vanilla array.

domoritz:
Arrow is already a dependency of duckdb wasm.

bitflower:
Cool, didn't know @domoritz 

Will come back with adjustments

Alex-Monahan:
Mind if we show both examples? I've seen cases where inserting as JSON worked, but converting to Arrow ahead of time failed (due to mismatched types within a column like 2 and '2')

Mytherin (approved):

PR #1284, opened 2023-06-12
Define `NDEBUG` in release mode

Y--:
Defines the `NDEBUG` macro in Release mode.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -71,6 +71,7 @@ endif ()
 if(EMSCRIPTEN)
   # Release build
   if(CMAKE_BUILD_TYPE STREQUAL "Release")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DNDEBUG")
     # ... with min size
     if(WASM_MIN_SIZE)
       set(CMAKE_CXX_FLAGS_RELEASE "-Oz")


carlopi:
Hi @Y--!

Could you potentially share where do you see this have an impact / rationale?
I assumed that CMake will already define this for any Release build, and it seems to be passed on at least to same components automatically, I will dig further but if you have anything to share it will be valuable (even just for future reference).

----

Locally with and without the commit I get this files for examples:
```
./scripts/wasm_build_lib.sh relperf eh #master
-rw-r--r--  1 carlo  staff     47628 Jun 12 23:36 build/relperf/eh/duckdb_wasm.js
-rwxr-xr-x  1 carlo  staff  17459015 Jun 12 23:36 build/relperf/eh/duckdb_wasm.wasm

./scripts/wasm_build_lib.sh relperf eh #master + NDEBUG added in Makefile
-rw-r--r--  1 carlo  staff     47393 Jun 12 23:50 build/relperf/eh/duckdb_wasm.js
-rwxr-xr-x  1 carlo  staff  17466782 Jun 12 23:50 build/relperf/eh/duckdb_wasm.wasm
```
These do not generates the exact same files, so there are changes to the setup.

---

And digging a bit further in the build folder there are differences, at least for the targets duckdb_wasm and duckdb_web (like in `build/relperf/eh/CMakeFiles/duckdb_wasm.dir/flags.make` where -DNDEBUG is not specified).

carlopi (approved):

carlopi:
Thanks for the contribution, LGTM.

Y--:
Hi @carlopi ! Thanks a lot for the quick review. I was surprised as well that `NDEBUG` was not set by CMake, but didn't get a chance to understand why exactly.
I've made sure it wasn't the case by gating an `#error` pragma behind `#ifdef NDEBUG` and trying to compile in release mode.

bleskes:
FYI https://stackoverflow.com/questions/34302265/does-cmake-build-type-release-imply-dndebug

@carlopi we have some assertion code under NDEBUG that slows things down considerably.

carlopi:
Hi @Y-- & @bleskes!

At first I was skeptical, but it's also fairly easy to check that you are right and -DNDEBUG is currently not passed on to `duckdb_wasm` and `duckdb_web` targets.

PR for me is ready to be merged, waiting for another set of eyes.

I think we (duckdb/duckdb-wasm) are potentially at fault and override the default behaviour in some way, I will try to track this down, mostly to verify also other options are passed on correctly to libraries (seems to be the case at the moment, but it's worth a deeper look).

Mytherin (approved):

PR #1292, opened 2023-06-16
Fix: handle s3 PUT silently ignoring HTTP errors

ozkatz:
S3 filesystem currently doesn't check the HTTP response code returned by S3 - trying to writing to e.g. missing bucket or with invalid credentials will fail silently.

This PR adds a simple HTTP status check - [PutObject](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html) on S3 is expected to return `200 OK` for successful writes.


carlopi:
Hi @ozkatz, thanks for the contribution, looks good to me!

samansmink (approved):
lgtm!

Mytherin (approved):

PR #1323, opened 2023-07-16
Multiple fixes, bump emscripten, DuckDB and arrows versions

carlopi:
Moving most jobs out of container for CI, final one have to be reviewed and also potentially moved over, but it proved a bit too complex, so going incrementally.

Bumped versions of Emscripten as tools, DuckDB and Arrows as dependencies.

Removed EXPORT_ES6 from Emscripten command line variables, given that introduced a generalization via generateRequire that was broken in node, harming both users and testing.


carlopi:
Would it be possible to merge this PR using either "rebase (no squash)" or "git merge"? Given that there are a few different lines of works, it would be nice to preserve the commits avoiding squash.

Mytherin:
Thanks!

Issue #1298, opened 2023-06-20
Error sort query with array of struct type


handstuyennn:
### What happens?

I built duckdb shell from latest code of duckdb-wasm.
I ran following queries and found got error `memory access out of bounds`:

- `CREATE TABLE  test (id integer PRIMARY KEY,struct_arr ROW(val1 integer, val2 varchar)[],);`
- `INSERT INTO test VALUES (1, [(1, 'test'), (2, 'lorem ipsum')]), (2, [(3, 'dolor'), (4, 'sit')])`
- `SELECT * FROM test ORDER BY id DESC;`

### To Reproduce
Start duckdb shell in browser with command `make app_start`
Run following query in duckdb wasm shell:

> CREATE TABLE  test (id integer PRIMARY KEY,struct_arr ROW(val1 integer, val2 varchar)[],);

> INSERT INTO test VALUES (1, [(1, 'test'), (2, 'lorem ipsum')]), (2, [(3, 'dolor'), (4, 'sit')]);

> SELECT * FROM test ORDER BY id DESC;

### Browser/Environment:

Chrome

### Device:

MacOS

### DuckDB-Wasm Version:

latest code from branch `master`

### DuckDB-Wasm Deployment:

http://localhost:9002/

### Full Name:

Tuyen Nguyen Ngoc

### Affiliation:

Developer

carlopi:
Hi @handstuyennn! I am not sure I understand what might be happening, since in theory you should get the same results between master and shell.

Could you try once again from a fresh folder following the instructions here: https://github.com/duckdb/duckdb-wasm#build-from-source?

handstuyennn:
Yes, I build exactly following the instructions in this file.
After that I ran command `make app_start`
And here is what I get in my browser.

![image](https://github.com/duckdb/duckdb-wasm/assets/49063786/cf605685-a3c5-4193-a8a1-879e4456ab0f)

carlopi:
Thanks for raising this, I do reproduce, I guess one big problem is that this path is not really tested, so this somehow slipped.

I will ping when I have fix, and I will make so that the default configuration should be tested regularly by CI (as to catch regression earlier).

carlopi:
[...]

PR #1328, opened 2023-07-17
Bump DuckDB, properly propagate wasm_* platform and add test

carlopi:
Previously the custom provided platform was not propagated in the right place, fixed and added a test.

Fixes https://github.com/duckdb/duckdb-wasm/issues/1298


Mytherin:
Thanks for the PR! Are all the `console.log` messages left in intentionally?

carlopi:
Yes, it's not great practice, but currently strategy of shallowing exceptions is not great since shallows both DuckDB-side exceptions (like failures) and JS exceptions (these might be harmless, and might make sense to skip in some cases).

To users not opening the console, it should give roughly the same experience, but upon inspection some information of what went wrong is available.
Unsure whether to consider this a blocker to tag a version or not, but we can check tomorrow.

Note that exceptions are still used only for exceptional behaviour, so I would assume (but I am not completely sure) that the number of exceptions will be low (bounded by the number of failed queries ?).

Mytherin:
Alright, makes sense to me. Thanks

whscullin:
Hi, just came across this when diagnosing log spamming during some of our tests. It would be nice to be able to disable the currently unconditional `console.log` so we can have clean test outputs. Working around this mocking `console.log` for now.

whscullin:
I guess it's unclear to me why this isn't just using the `Logger` interface, at least in the worker where it's available. that seems like it would allow for configuration. I can do a PR for that if that seems reasonable.

carlopi:
I moved this to an independent issue, #1518.

Why this happened is an oversight on my side, there was a comment like:
```cpp
                    // Next few lines should argubly in separate JavaScript-land function call
                    // TODO: move them out / have them configurable
```
that I never got to address.

Fix is trivial, only problem is that this is in DuckDB codebase and will require patching while building duckdb in duckdb-wasm, but there is already infrastructure for that.

PR #1330, opened 2023-07-19
Benchmarks: Restore to functional state & execute benchmarks on CI

carlopi:
This PR deals only with performing the micro benchmarks, but I have a follow up PR where I would like to re-enabled back also the TPCH benchmarks.


carlopi:
I think this is good and can land.
This PR restores support for benchmarks and have them executed as an additional step as part of the main workflow (only when on master).

Benchmarks results are uploaded both as arrow file (as was done before, to be used by the versus page), but also as json and csv files.

ankoh (approved):

Mytherin:
Thanks!

PR #1338, opened 2023-07-24
Don't run test in building targets

Y--:
As client of this library, we want to build it but not necessarily run the tests each time.

This does the following changes:
* remove test execution from `js_debug` and `js_release` building targets
* add `js_tests_release` to run tests on release build
* don't change `js_tests` which depends on `js_debug` target already

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -344,7 +344,7 @@ app_start_corp:
 	yarn workspace @duckdb/duckdb-wasm-app start:corp
 
 .PHONY: app
-app: wasm wasmpack shell docs
+app: wasm wasmpack shell docs js_tests_release
 	yarn workspace @duckdb/duckdb-wasm-app build:release
 
 .PHONY: app_server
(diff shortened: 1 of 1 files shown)


Y--:
> Great, thanks for the feedback.
> 
> I would consider adding `js_tests_release` to be a descendent of `app` target, directly or indirectly, so that the basic exprience of cloning and running `make` should run some tests, or do you consider that also noisy? Feel free to add it or I can do it in a follow up.

Yeah I think running tests makes sense in this case. Added [here](https://github.com/duckdb/duckdb-wasm/pull/1338/commits/75ca01e1a85a26d24e666cebe785e8aff4ce73ed), if I understood correctly :-)

carlopi:
> Yeah I think running tests makes sense in this case. Added [here](https://github.com/duckdb/duckdb-wasm/pull/1338/commits/75ca01e1a85a26d24e666cebe785e8aff4ce73ed), if I understood correctly :-)

I guess so, thanks for the fix.

For me it's ready to go (eslint error is random)

PR #1349, opened 2023-08-04
Adapt to ArrowOptions removal

carlopi:


carlopi:
I took a shortcut in re-introducing (via `duckdb.patch`) support for serialising TIME_TZ to arrow, that was discontinued in https://github.com/duckdb/duckdb/pull/8089.

For now just moving forward, but it will be to have properly looked into given arrow interface role in duckdb-wasm.

Mytherin:
Thanks!

PR #1369, opened 2023-08-17
Move to main as default branch

carlopi:
This is to be merged just after changing master to main as default branch for duckdb-wasm


carlopi:
Failure is expected, this has to merged after moving master to main as default branch

PR #1396, opened 2023-08-31
upgrade to apache-arrow v13

jraymakers:
Upgrade apache-arrow, in part to get the fix for this bug: https://github.com/apache/arrow/pull/35067

### DIFF
--- a/packages/benchmarks/package.json
+++ b/packages/benchmarks/package.json
@@ -5,7 +5,7 @@
     "private": true,
     "dependencies": {
         "@duckdb/duckdb-wasm": "file:../duckdb-wasm",
-        "apache-arrow": "^11.0.0",
+        "apache-arrow": "^13.0.0",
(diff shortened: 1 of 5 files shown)


carlopi:
Thanks for the PR, I need to take a deeper look at this to check if there was any specific reason to stay with older arrow, or was just not updated by accident.

Mytherin:
Thanks!

jraymakers:
> Thanks for the PR, I need to take a deeper look at this to check if there was any specific reason to stay with older arrow, or was just not updated by accident.

This version (v13) was only [released](https://www.npmjs.com/package/apache-arrow?activeTab=versions) a week or two ago. The [last time the version was changed](https://github.com/duckdb/duckdb-wasm/commit/9e8ff405837a30d5f0cb16d5d3224e4e4176495f) (to v12) was before v13 was released.

PR #1402, opened 2023-09-05
Use Puppeteer for Chrome dependency in Karma tests

dude0001:
The purpose of this change is to improve onboarding for new developers. Currently running `make` requires Chrome to be installed. If Chrome is not installed and configured in the `CHROME_BIN` environment variable, Karma tests fail with an error like:

```
ERROR [launcher]: No binary for ChromeHeadless browser on your platform.
  Please, set "CHROME_BIN" env variable.
```

The Puppeteer [getting started](https://github.com/puppeteer/puppeteer#getting-started) documentation states "When you install Puppeteer, it automatically downloads a recent version of [Chrome for Testing](https://goo.gle/chrome-for-testing) (~170MB macOS, ~282MB Linux, ~280MB Windows) that is [guaranteed to work](https://pptr.dev/faq#q-why-doesnt-puppeteer-vxxx-work-with-chromium-vyyy) with Puppeteer. The browser is downloaded to the $HOME/.cache/puppeteer folder by default...".

With this change users that do not have Chrome installed and configured in `CHROME_BIN` will instead default to using the version of Chrome installed as a dev package dependency with Puppeteer.

```
> @duckdb/duckdb-wasm@1.11.0 test:chrome
> karma start ./karma/tests-chrome.cjs

CHROME_BIN=/home/dude0001/.cache/puppeteer/chrome/linux-116.0.5845.96/chrome-linux64/chrome
....
05 09 2023 10:47:28.830:INFO [karma-server]: Karma v6.4.2 server started at http://localhost:9876/
```

Puppeteer has some other functionality that could eventually be used. Currently, it would only be used to install the dependency on Chrome as a package instead of a prerequisite. If desired, a developer and the CI pipeline can configure `CHROME_BIN` to use a specific version of Chrome.

### DIFF
--- a/packages/duckdb-wasm/karma/karma.base.cjs
+++ b/packages/duckdb-wasm/karma/karma.base.cjs
@@ -1,3 +1,7 @@
+if (!process.env.CHROME_BIN) {
+    process.env.CHROME_BIN = require('puppeteer').executablePath();
+}
+
 const JS_TIMEOUT = 900000;
 
 module.exports = function (config) {
--- a/packages/duckdb-wasm/package.json
+++ b/packages/duckdb-wasm/package.json
@@ -50,7 +50,7 @@
         "make-dir": "^4.0.0",
         "nyc": "^15.1.0",
         "prettier": "^3.0.2",
-        "puppeteer": "^20.8.0",
+        "puppeteer": "^21.1.1",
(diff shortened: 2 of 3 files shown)


carlopi (approved):
Thanks a lot for the contribution, keep them coming!

PR #1403, opened 2023-09-07
Loadable extensions

carlopi:
duckdb.patch is the diff actually in https://github.com/duckdb/duckdb/pull/8827, so might make more sense for that to be merged first, and then removing the duckdb.patch file AND bumping submodule duckdb to the relevant commit.

---

Very much "Works on my machine" vibes, and this is worth of a proper explainer, but...
```
git clone https://github.com/carlopi/duckdb-wasm.git
git checkout loadable-extensions
cd duckdb-wasm
git submodule init
git submodule update
cd submodules/duckdb
git apply ../../duckdb.patch
cd ../../
make clean && make build_loadable && make serve_loadable
```
Should launch a duckdb-wasm shell on your browser, where typing:
```
duckdb> SET custom_extension_repository='http://127.0.0.1:8081';
duckdb> SUMMARIZE FROM 'https://shell.duckdb.org/data/benchmarks.json';
```
Will autoload json extension, and show some summary information for that json file.


carlopi:
Almost there, current failures are due to:
```
Error: Extension Autoloading Error: An error occured while trying to automatically install the required extension 'parquet':
Extension "http://extensions.duckdb.org/duckdb-wasm/d5c4422f72/wasm_eh/parquet.duckdb_extension.wasm" not found.
Extension "parquet" is an existing extension.

Install it first using "INSTALL parquet".
```

PR #1410, opened 2023-09-13
Loadable extensions: deploying on shell.duckdb.org

carlopi:
Two commits that would need some polishing, but I think we should be able to go live with this.


Mytherin:
Awesome!

PR #1418, opened 2023-09-24
Add hint on how to get query result as JSON

tobilg:
This will eventually make it easier for people to understand hot to transform the Arrow Table to JSON. As the [Apache Arrow JS Docs](https://arrow.apache.org/docs/js/index.html) as empty, this is not really easy to find.

It's also aligned to the main DuckDB docs at https://duckdb.org/docs/api/wasm/query#arrow-table-to-json

### DIFF
--- a/packages/duckdb-wasm/README.md
+++ b/packages/duckdb-wasm/README.md
@@ -210,6 +210,11 @@ for await (const batch of await conn.send<{ v: arrow.Int }>(`
 `)) {
     // ...
 }
+// Get query results as JSON
+const result = await conn.query<{ v: arrow.Int }>(`
+    SELECT * FROM generate_series(1, 100) t(v)
+`);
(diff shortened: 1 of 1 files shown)


domoritz:
I'd not recommend doing this conversion unless absolutely necessary. It incurs a lot of overhead and arrow table as already behave like arrays of objects. 

Thanks for pointing at the missing arrow docs. I'll look into it.

tobilg:
My use case (and I believe a lot of other people‘s as well) is to show the data in a React application as table. I‘m not aware of any components/libraries that can handle arrow table data directly. 

I‘m happy to learn about other ways to do this. 

Also, as written, this is covered in the DuckDB docs.

bluehat974:
@domoritz Issue created
https://github.com/apache/arrow/issues/37856

domoritz:
And I filed https://github.com/apache/arrow/issues/37858 about the docs missing.

PR #1422, opened 2023-09-26
Dependabots

carlopi:


Mause:
@carlopi can you add this config? https://github.blog/2023-08-24-a-faster-way-to-manage-version-updates-with-dependabot/

It should make dependabot raise multiple versions bumps at once in a single pr

carlopi:
There is also something off in the CI configuration, so dependabots PRs run twice: https://github.com/duckdb/duckdb-wasm/pull/1415, once failing and once passing.

Always a puzzle.

carlopi:
> @carlopi can you add this config? https://github.blog/2023-08-24-a-faster-way-to-manage-version-updates-with-dependabot/
> 
> It should make dependabot raise multiple versions bumps at once in a single pr

Thanks, makes sense, will take a look at this.

PR #1424, opened 2023-09-26
esbuild-browser example using coi

jraymakers:
Add the COI build to the esbuild-browser example.

In order for this to work, we need to serve the page with the COI headers (COEP & COOP). Since it doesn't appear that `http-server` supports custom headers, I added an alternate HTTP server, `serve`, along with configuration to set these headers. I've left `http-server` in place, so the non-COI case can still be tested easily.

I've tested this in Chrome, Firefox, and Safari on an M2 Mac (latest versions of all browsers & OS), and it works fine with the basic query in the example. I have not tested other queries. In Chrome and Firefox, I do see multiple (4) network requests for the pthread worker script, which is evidence that multiple threads are being started. I don't see this in Safari; I don't know whether that means multiple threads are not being used, or just that Safari reports this information differently.

### DIFF
--- a/examples/esbuild-browser/.gitignore
+++ b/examples/esbuild-browser/.gitignore
@@ -1,3 +1,4 @@
 /node_modules
 /*.wasm
 /*.js
+/*.js.map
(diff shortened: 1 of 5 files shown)


jraymakers:
This was originally just a demonstration PR, but I'd like to actually merge it.
It provides a simple working example of how to use the COI build. It works on all major browsers.
It only changes one example, and it's purely additive.

carlopi:
I did not manage to have this properly work on my setup, I will give it another try, but it seems this is just moving the problem a bit further.

jraymakers:
> I did not manage to have this properly work on my setup, I will give it another try, but it seems this is just moving the problem a bit further.

I'm interested to hear what problems you encounter. For reference, I'm using an M2 Mac and testing in Chrome.

If there are problems with COI in some environments, then an example like this could make them easier to discover and diagnose.

carlopi:
I think this can be closed in favor of #1501, thanks a lot @jraymakers for bringing this in.

PR #1434, opened 2023-10-06
expose web file type

jraymakers:
Export the `WebFile` type from the client library.

This type is returned from the exported method `globFiles` (on `AsyncDuckDB`), but the type itself is not exported, so using that method is challenging in some contexts.


carlopi (approved):
Thanks for contributing this change!

carlopi (approved):
Thanks for contributing this change!

PR #1468, opened 2023-11-01
Issue 1467: Add test that this works in regular duckdb-wasm

carlopi:
Issue to stay open since it's still a problem in the Web shell


arroyo-cel:
> Problema que debe permanecer abierto ya que todavía es un problema en el shell web.

Starjimena10@gmail.com

PR #1474, opened 2023-11-08
Link embind

Y--:
Emscripten necessitate that you link all libraries with the main module in order for the dynamically-loaded modules to use them.

So in order to use [Embind](https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html) tooling in a DuckDB loadable extension, we need to link `embind` with the main module.

This is what this PR enable.


carlopi (approved):

hannes:
Thanks!

PR #1475, opened 2023-11-08
Expose `DUCKDB_BINDINGS` in global scope

Y--:
To let DuckDB extensions access the main module, we would need to have an access to the `DuckDBBindings` somehow.

This PR exposes them under `DUCKDB_BINDINGS` (to be consistent with `DUCKDB_RUNTIME` that is also exposed)

Let me know if you think this exposes too much, I can try to think of other way to achieve this too, though this would be by far the simplest and most flexible solution :-)


carlopi (approved):
Thanks!

PR #1476, opened 2023-11-08
Don't minify debug builds

Y--:
A small change that I found myself doing quite often locally so I figured I would suggest it :-)


carlopi (approved):
Thanks a lot!

PR #1491, opened 2023-11-16
fix insertCSVFromPath in README

candux:
insertCSVFromPath() is a function of the connection and not the database.
Error message:
TypeError: Cannot read properties of undefined (reading 'columns')


carlopi:
This should ideally be properly tested so not to get out-of-sync and avoid this similar problems in the future, but improving this is already nice, thanks a lot for filing the PR.

carlopi (approved):

PR #1493, opened 2023-11-17
Fix use of insertCSVFromPath

szarnyasg:
Fix issue pointed out in https://github.com/duckdb/duckdb-web/pull/1548


candux:
I already created an MR for this which also fixes another issue https://github.com/duckdb/duckdb-wasm/pull/1491

szarnyasg:
Ah, thanks for pointing that out. I missed that.

PR #1490, opened 2023-11-16
Add OPFS support

dengkunli:
This PR adds [OPFS (Origin Private File System)](https://web.dev/articles/origin-private-file-system) support for duckdb-wasm. With OPFS support, we could persist the database in the browser 🔥

### API

To remain simple and intuitive, no new api is added, just pass a `path` starting with `opfs://` to `db.open` and we'll know we're using OPFS and finish all the preparation work needed:

```js
await db.open({
    path: 'opfs://test.db',
    accessMode: duckdb.DuckDBAccessMode.READ_WRITE,
});
const conn = await db.connect();
```

### List of Changes

* Add OPFS Support
* Improve Debuggability: use `-g` option to enable all DWARF debugging capabilities (before this, we could breakpoint but debug symbol information are missing)
* Fix the coi worker build: the `duckdb-browser-coi.pthread.worker.ts` override the global onmessage handler but forgot to set `startWorker` in the `load` cmd (when missing, the coi worker will not start properly)
* Update examples/esbuild-browser: add opfs example & support serving coi build
* Add tests

PR #1501, opened 2023-11-20
Example using coi: thanks @jraymakers & @dengkunli

carlopi:
There were two independent PRs on enabling examples using COI:
* https://github.com/duckdb/duckdb-wasm/pull/1424, that was missing a bit of the pthread syncronization dance
* https://github.com/duckdb/duckdb-wasm/pull/1490, that is centered around OPFS, and this seems an orthogonal feature

I took the liberty of merging them into this PR.

Thank you both for the contribution.

I will add enabling shell as a follow up.

### DIFF
--- a/examples/esbuild-browser/.gitignore
+++ b/examples/esbuild-browser/.gitignore
@@ -1,3 +1,4 @@
 /node_modules
 /*.wasm
 /*.js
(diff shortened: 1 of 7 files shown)


Mytherin:
Thanks

Issue #1502, opened 2023-11-20
Error when query contains a UNION value


jraymakers:
### What happens?

When the following query is run in DuckDB WASM:

`from (values (union_value(str := 'abc')));`

an error like the following is returned:

```
Cannot destructure property 'length' of '(intermediate value)(intermediate value)(intermediate value)' as it is undefined.
    at VectorLoader.readData (vectorloader.ts:125:56)
    at VectorLoader.visitUtf8 (vectorloader.ts:70:151)
    at VectorLoader.visit (visitor.ts:26:48)
    at VectorLoader.visit (vectorloader.ts:54:22)
    at visitor.ts:23:44
    at Array.map (<anonymous>)
    at VectorLoader.visitMany (visitor.ts:23:22)
    at VectorLoader.visitSparseUnion (vectorloader.ts:103:150)
    at VectorLoader.visit (visitor.ts:26:48)
    at VectorLoader.visit (vectorloader.ts:54:22)
```

This happens when the query is passed directly to the DuckDB WASM JS API (i.e. `query` on `AsyncDuckDBConnection`).

In `shell.duckdb.org`, running the query above does not return any results, either an error or the correct output.

In the native CLI, the query above returns:

```
D from (values (union_value(str := 'abc')));
┌────────────────────┐
│        col0        │
│ union(str varchar) │
├────────────────────┤
│ abc                │
└────────────────────┘
```

### To Reproduce

In the `esbuild-browser` example, replace the query with `from (values (union_value(str := 'abc')));`.

### Browser/Environment:

Chrome 118

### Device:

M2 MacBook

### DuckDB-Wasm Version:

1.28.0

### DuckDB-Wasm Deployment:

local (esbuild-browser example)

### Full Name:

Jeff Raymakers

### Affiliation:

MotherDuck

jraymakers:
This does not appear to depend on the type of the value in the UNION. For example, using an integer results in the same error with a slightly different stack trace:

`from (values (union_value(num := 123)));`

```
Cannot destructure property 'length' of '(intermediate value)(intermediate value)(intermediate value)' as it is undefined.
    at VectorLoader.readData (vectorloader.ts:125:56)
    at VectorLoader.visitInt (vectorloader.ts:64:113)
    at VectorLoader.visit (visitor.ts:26:48)
    at VectorLoader.visit (vectorloader.ts:54:22)
    at visitor.ts:23:44
    at Array.map (<anonymous>)
[...]

PR #1503, opened 2023-11-23
upgrade apache-arrow to v14.0.1

jraymakers:
Upgrade apache-arrow from v13.0.0 to v14.0.1. This should fix [this issue](https://github.com/duckdb/duckdb-wasm/issues/1502).

### DIFF
--- a/packages/benchmarks/package.json
+++ b/packages/benchmarks/package.json
@@ -5,7 +5,7 @@
     "private": true,
     "dependencies": {
         "@duckdb/duckdb-wasm": "file:../duckdb-wasm",
-        "apache-arrow": "^13.0.0",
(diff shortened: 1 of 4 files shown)


carlopi (approved):

pmm-motif:
BTW, I'm wondering if this might fix this issue as well: https://github.com/duckdb/duckdb-wasm/issues/1477

Trying to build duckdb-wasm locally and test but it's not that easy 😅

carlopi:
I don't think so, that looks like a problem in the duckdb-wasm "frontend", I will have to go back to that.

ankrgyl:
@carlopi is this something we can help with? Do you have a sense of whether a change in duckdb-wasm or duckdb caused the issue?

PR #1527, opened 2023-12-11
Output extension to a sub-directory that respect the input mode/feature

Y--:
Today, the `build_loadable.sh` scripts output every extension under `loadable_extensions` folder.

This PR outputs them under `loadable_extensions/$MODE/$FEATURE` where `MODE` and `FEATURE` are the parameters provided in input to select the source directory (eg. `mode=relperf feature=mvp`)

### DIFF
--- a/scripts/build_loadable.sh
+++ b/scripts/build_loadable.sh
@@ -4,16 +4,17 @@ set -euo pipefail
 
 trap exit SIGINT
 
-PROJECT_ROOT="$(cd $(dirname "$BASH_SOURCE[0]") && cd .. && pwd)" &> /dev/null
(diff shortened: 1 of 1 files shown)


carlopi (approved):
I though about and solving this generally (= in duckdb Makefile), with having a postprocessing step where extensions are copied to `build/release/extensions/${duckdb_version}/${platform}/${name}.duckdb_extension[.wasm]`, since that will also allow to use the folder as a local extension directory, basically `SET custom_extension_repository='/path/to/folder';`

Stuff might move more that direction, but for not this is still more clean.

Thanks.

PR #1530, opened 2023-12-14
Fix for emscripten 3.1.51

Y--:
The latest Emscripten version ([3.1.51](https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md#3151---121323)) seem to have broken the build with:

```
Could not resolve "vm"

    src/bindings/duckdb-coi.pthread.js:1:357:
      1 │ ...=require("fs");var vm=require("vm");Object.assign(global,{self:g...
        ╵                                  ~~~~

  The package "vm" wasn't found on the file system but is built into node. Are you trying to bundle for node? You can use "platform: 'node'" to do that, which will remove this error.

/home/runner/work/mono/mono/hatchling/client/wasm/duckdb-wasm/node_modules/esbuild/lib/main.js:1650
  let error = new Error(text);
              ^

Error: Build failed with 1 error:
src/bindings/duckdb-coi.pthread.js:1:357: ERROR: Could not resolve "vm"
    at failureErrorWithLog (/home/runner/work/mono/mono/hatchling/client/wasm/duckdb-wasm/node_modules/esbuild/lib/main.js:1650:15)
    at /home/runner/work/mono/mono/hatchling/client/wasm/duckdb-wasm/node_modules/esbuild/lib/main.js:1059:25
    at /home/runner/work/mono/mono/hatchling/client/wasm/duckdb-wasm/node_modules/esbuild/lib/main.js:1004:52
    at buildResponseToResult (/home/runner/work/mono/mono/hatchling/client/wasm/duckdb-wasm/node_modules/esbuild/lib/main.js:1057:7)
    at /home/runner/work/mono/mono/hatchling/client/wasm/duckdb-wasm/node_modules/esbuild/lib/main.js:1086:16
    at responseCallbacks.<computed> (/home/runner/work/mono/mono/hatchling/client/wasm/duckdb-wasm/node_modules/esbuild/lib/main.js:703:9)
    at handleIncomingPacket (/home/runner/work/mono/mono/hatchling/client/wasm/duckdb-wasm/node_modules/esbuild/lib/main.js:763:9)
    at Socket.readFromStdout (/home/runner/work/mono/mono/hatchling/client/wasm/duckdb-wasm/node_modules/esbuild/lib/main.js:679:7)
    at Socket.emit (node:events:514:28)
    at addChunk (node:internal/streams/readable:5[45](https://github.com/motherduckdb/mono/actions/runs/7208649622/job/19639532649#step:13:46):12) {
  errors: [Getter/Setter],
  warnings: [Getter/Setter]
}
```

cf. this [build](https://github.com/duckdb/duckdb-wasm/actions/runs/7210471199/job/19644493777?pr=1530)

This fails is because ESBuild attempts to bundle the NodeJS package `vm`. The proposed fix is to make it invisible to ESBuild by replacing `require('vm')` with `['vm'].map(require)`, which ESBuild doesn't statically analyze and replace.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,7 @@ compile_commands.json
 data/tpch/**/*.parquet
 .idea
 lib/.idea
+
+check_duckdb
+wasm_setup
+loadable_extensions/*.wasm
--- a/packages/duckdb-wasm/bundle.mjs
+++ b/packages/duckdb-wasm/bundle.mjs
@@ -4,6 +4,7 @@ import path from 'path';
 import { rimrafSync } from 'rimraf';
 import mkdir from 'make-dir';
 import { fileURLToPath } from 'url';
+import { execSync } from 'child_process';
 
 // -------------------------------
 // Current bundling strategy
@@ -195,6 +196,10 @@ fs.copyFile(path.resolve(src, 'bindings', 'duckdb-coi.wasm'), path.resolve(dist,
     });
 
     console.log('[ ESBUILD ] duckdb-browser-coi.worker.js');
+    // Don't attempt to bundle NodeJS modules in the browser build.
+    execSync(
+        `sed -i.bak 's/require("vm")/["vm"].map(require)/g' ./src/bindings/duckdb-coi.pthread.js && rm ./src/bindings/duckdb-coi.pthread.js.bak`,
+    );
     await esbuild.build({
         entryPoints: ['./src/targets/duckdb-browser-coi.worker.ts'],
         outfile: 'dist/duckdb-browser-coi.worker.js',


carlopi:
Ouch, not a great news, thanks for taking on this

carlopi:
Haven't looked into this yet, so depends, but one way out, even just temporary, might be saying that 3.51 is not supported, skip it in CI and move further.

carlopi:
One question I have is whether that will hold in a browser, where the module can't be found, and that is currently not tested by the CI run.

I would be still for skipping this version for now, or provide two versions of the threaded worker (one for use in node and one in browser), and see how they are different.

I have updated and I am compiling locally, back in a bi.

Y--:
> One question I have is whether that will hold in a browser, where the module can't be found, and that is currently not tested by the CI run.

There's no problem in the browser because the `require('vm')` is gated with a test on `process.versions.node` which is only true in NodeJS:
image

The reason why ESBuild fails is because it does static analysis of the code, so it doesn't evaluate that condition.

carlopi (approved):

carlopi:
Just to elaborate on this, I think both problem and solution are not great to deal with, solution is potentially brittle to tooling becoming just a bit smarter (but not properly smart to recognise the if (IS_NODE) ).

But then, this is the technology stack, we will deal with it. This can be merged and moving on.

Thanks @Y-- for taking care of this.

PR #1546, opened 2023-12-22
Patch all bindings

Y--:
Now that `node` is an Emscripten environment, we need to patch all JS bindings to be able to run ESbuild bundling with DuckDB WASM build with loadable extension mode.


carlopi (approved):
Thanks!

PR #1547, opened 2023-12-23
fix unit tests

jraymakers:
A few unit tests were consistently failing for me locally. This should fix them.

The all_types test was missing the "union" type, which was recently added to `test_all_types()`.
(The other changes to this file are automatic formatting fixes applied by VS Code according to the settings in the repo.)

A few tests relied on generated TPCH files. The generation script was failing because it didn't load the (newly loadable) `tpch` extension. I added the appropriate "install" and "load" commands to this script.

### DIFF
--- a/duckdb.patch
+++ b/duckdb.patch
@@ -1,23 +1,3 @@
-diff --git a/src/function/table/system/test_all_types.cpp b/src/function/table/system/test_all_types.cpp
-index a0b856266c..753fd6323b 100644
---- a/src/function/table/system/test_all_types.cpp
-+++ b/src/function/table/system/test_all_types.cpp
-@@ -204,13 +204,14 @@ vector<TestType> TestAllTypesFun::GetTestTypes(bool use_large_enum) {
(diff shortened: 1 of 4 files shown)


jraymakers:
Strangely, without this change, the all_types test fails in my local environment with `Expected 44 to equal 43.`, but with this change, the GitHub CI build fails with `Expected 43 to equal 44.` I don't understand why the expected numbers are different.

jraymakers:
Oh, it looks like the CI applies a patch to duckdb (duckdb.patch), and part of that patch is to remove the `union` column from `test_all_types()`. This patch is not applied in a local build.

jraymakers:
I've removed the patch that removed `union` from `test_all_types` in the CI. This caused the C++ `all_types_test` to fail, since it didn't have a test for `union`, so I added that.

Since the published version has `union`, this should bring the CI version closer to the published version.

carlopi (approved):
Thanks a lot for the PR, makes a lot of sense to me.

Mytherin:
Thanks!

PR #1572, opened 2024-01-08
Handle single quotes

Y--:
In some circumstances the generated code contains single quotes (`require('vm')` rather than `require("vm")`. This change handles that case.


carlopi (approved):

Mytherin:
Thanks!

PR #1573, opened 2024-01-08
Allow to specify bundle

Y--:
While developing, I found it convenient to be able to switch from one bundle to another using the `bundle` query parameter, so I thought I would contribute it.

It can be used in the Shell like this for example:
http://127.0.0.1:8081/?bundle=mvp
http://127.0.0.1:8081/?bundle=eh

Happy to adjust the implementation if need be.


carlopi (approved):
Thanks, this makes a lot of sense.

Mytherin:
Thanks!

carlopi:
And it's live: https://shell.duckdb.org/?bundle=eh or https://shell.duckdb.org/?bundle=mvp, do work, great!

PR #1589, opened 2024-01-17
Clean-up JavaScript from unneeded exported functions

carlopi:
This drastically reduce the size of the generated JavaScript worker code, from 45M uncompressed to 2M, and reduces also the time spent processing the JS file and creating a bunch of (unnecessary) wrappers.

Problem is connected to dynamic loading of extensions, since I haven't found a proper way to express "symbol should be available to other dlopen-ed modules BUT it's not needed on the JavaScript side".

I am not sure whether this is a current Emscripten limitation or if the compiler wrapper should be invoked in a different way.

Current solution is blunt and exploit current Emscripten specifics code-generation around `var wasmExports = createWasm();`. This needs to be reworked to be compatible with future versions.

Basically before this PR for any symbol exported at the C++ level a function like:
```
var __function_name = Module["__function_name"] = (a0) => (__function_name = Module["__function_name"] = wasmExports["_function_name"])(a0);
```
would be generated.

Multiplied 80K symbols, it meant a lot of JavaScript code had to be downloaded, parsed and executed for no actual use.

This should solve https://github.com/duckdb/duckdb-wasm/issues/1561, but I will have to properly check after this gets merged.

This adds a (compile-time) dependency on awk (that should be widely available) and on the npm module js-beautify.

PR #1590, opened 2024-01-17
Clean-up JavaScript from unneeded exported globals

carlopi:
Follow up of https://github.com/duckdb/duckdb-wasm/pull/1589, now avoiding more unnecessary code by filtering for exported C++ globals to JavaScript.

duckdb-worker-eh.js goes from 2.0M to 1.2M.

Also add attempt at measuring package size via pkg-size npm package.

Then using your favourite DBMS:
```
npx pkg-size --sizes=size --json > pkg-sizes.json
duckdb -c "select sum (size) from (select unnest(f) from (select unnest(files) as f from read_json_auto('pkg-sizes.json')));"
duckdb -c "select sum (size), sum(size) < '150000000' from (select unnest(f) from (select unnest(files) as f from read_json_auto('pkg-sizes.json')));"
```
would allow to have the measurement summed up.

There has to be a more proper way of measuring sizes by first building a local package, installing it elsewhere and checking the content, I will have to look into that.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1049,6 +1049,10 @@ jobs:
                   name: duckdb-wasm-packages-loadable
                   path: duckdb-wasm-packages.zip
 
+            - name: Measure NPM stats
+              run: |
+                  ./scripts/npm_measure_lib.sh
+
             - name: Publish to NPM
               env:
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks!

PR #1601, opened 2024-01-22
Extensions: Optimize build to remove most exports via SIDE_MODULE=2

carlopi:
Improvement depends a lot on the extensions, can be 0-30%. For example, icu.duckdb_extension.wasm goes from 8.0MB to 6.3MB (uncompressed size)


carlopi:
This is the linked PR in spatial: https://github.com/duckdb/duckdb_spatial/pull/234

carlopi:
Also adding here, to be addressed in a future PR or at the Emscripten or even standardization level:
duckdb-wasm extensions do not expose their symbol at the JavaScript level, so the exports do not need to have the actual mangled names as export names, but they could have unnamed export names.

PR #1614, opened 2024-01-30
DO NOT MERGE, this is a test, expected to fail

carlopi:
Check if the 150MB size detection actually works


carlopi:
Test failed as expected, so detection works.

PR #1619, opened 2024-02-02
More proper formatting of the shell

carlopi:
Code is rough / to be refined.


carlopi:
#1629 implements this properly

PR #1629, opened 2024-02-08
Proper formatting / highlighting for queries URL encoded

carlopi:


Mytherin:
Thanks!

Mause:
Any chance of tests or `cargo fmt`?

carlopi:
> Any chance of tests or `cargo fmt`?

Tests for shell behaviour would be nice, but there are many unknowns and I don't have a proper idea of the infrastructure / it will end up being very finnicky.

`cargo fmt` I should add

PR #1635, opened 2024-02-13
Added plain html example

dberardo-com:
Closes https://github.com/duckdb/duckdb-wasm/discussions/1632

### DIFF
--- a/examples/plain-html/index.html
+++ b/examples/plain-html/index.html
@@ -1,0 +1,51 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Document</title>
+  </head>
+  <body>
+    <script>
+      const getDb = async () => {
+        const duckdb = window.duckdbduckdbWasm;
+        // @ts-ignore
+        if (window._db) return window._db;
+        const JSDELIVR_BUNDLES = duckdb.getJsDelivrBundles();
+
+        // Select a bundle based on browser checks
+        const bundle = await duckdb.selectBundle(JSDELIVR_BUNDLES);
+
+        const worker_url = URL.createObjectURL(
+          new Blob([`importScripts("${bundle.mainWorker}");`], {
+            type: "text/javascript",
+          })
+        );
+
+        // Instantiate the asynchronus version of DuckDB-wasm
+        const worker = new Worker(worker_url);
+        // const logger = null //new duckdb.ConsoleLogger();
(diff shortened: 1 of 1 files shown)


carlopi (approved):
Thanks a lot!

examples/plain-html/index.html
```diff
@@ -0,0 +1,51 @@
...
+        const db = new duckdb.AsyncDuckDB(logger, worker);
+        await db.instantiate(bundle.mainModule, bundle.pthreadWorker);
+        URL.revokeObjectURL(worker_url);
+        window._db = db;
+        return db;
+      };
+    </script>
+    <script type="module">
+      import * as duckdbduckdbWasm from "https://cdn.jsdelivr.net/npm/@duckdb/duckdb-wasm@1.28.1-dev106.0/+esm";
+      window.duckdbduckdbWasm = duckdbduckdbWasm;
```
WebReflection, line 38: why is this needed at all? there is a module system, you can use it right away. You also have top-level await in there so I hope this example can be improved. If you need help, I'd be happy to help with such improvement.

WebReflection: A better example that doesn't need to leak globally ... it's just an ECMAScript module, check duckdb-loader.js: https://pyscript.com/@agiammarchi/jolly-silence-copy/latest

dberardo-com: what happens if 2 different files import the same duckdb-loader. would duckdb be initialized twice ? or would the PackageQueryApp be singleton ?

WebReflection: in ESM you can import the same thing dozen times its code will run only once and its exports will be available to any other importer of the same file ... if that's what you are asking. If it's not, your "*trashing as global reference*" is not a solution, there's not even a check if it was there before, so I am not sure I am following the question or the concern, but nothing is solved better (or at all) in the current state of affairs.

WebReflection: it depends ... if you want a single connection create a module that returns always the same connection

dberardo-com: that's actually why i am using window._db in the script, to store the connection. is this equivalent to what you are suggesting? not sure how to store state in a module

WebReflection:
```js
import * as duckdbduckdbWasm from "whatever";
let _db;
export default (...args) => _db || (
  _db = duckdbduckdbWasm.connect(...args)
);
```

you have that module as `./duckdb-singleton.js` and you're done?

WebReflection: I did update my previous answer expanding on various things, sorry for the extra ping.

PR #1636, opened 2024-02-14
Bump to DuckDB v0.10.0 [ DO NOT MERGE YET ]

carlopi:
Required changes to enable to compile duckdb-wasm with DuckDB v0.10.0.

This is NOT in a mergeable state, but I think it could be worth having out.

There is still the duckdb.patch, that needs to be applied on top.

I am aware of two failures (CSV and Parquet writing, unsure what changed duckdb-side that could touch this), so this is not mergeable as-is, I will update the PR with the fix later.

Note that UHUGEINTs are not yet handled on the Arrow side, so currently removed from the tests.

As part of this I also removed some unnecessary CI / infrastructure, but I will also make another pass on this.


Makefile
```diff
@@ -361,12 +361,12 @@ app: wasm wasmpack shell docs js_tests_release
...
 build_loadable:
 	cp .github/config/extension_config_wasm.cmake submodules/duckdb/extension/extension_config.cmake
-	DUCKDB_WASM_LOADABLE_EXTENSIONS="signed" GEN=ninja ./scripts/wasm_build_lib.sh relsize eh
+	DUCKDB_PLATFORM="wasm_eh" DUCKDB_WASM_LOADABLE_EXTENSIONS="signed" GEN=ninja ./scripts/wasm_build_lib.sh relsize eh
 	bash ./scripts/build_loadable.sh relsize eh
 
 build_loadable_unsigned:
 	cp .github/config/extension_config_wasm.cmake submodules/duckdb/extension/extension_config.cmake
-	DUCKDB_WASM_LOADABLE_EXTENSIONS="unsigned" GEN=ninja ./scripts/wasm_build_lib.sh relsize eh
+	DUCKDB_PLATFORM="wasm_eh" DUCKDB_WASM_LOADABLE_EXTENSIONS="unsigned" GEN=ninja ./scripts/wasm_build_lib.sh relsize eh
```
Y--, line 369: Maybe we could `export DUCKDB_PLATFORM="wasm_${FEATURES}"` in `./scripts/wasm_build_lib.sh` so it is not required when calling the script?

carlopi:
Need a couple of more things in, reopening after cleaning this up.

PR #1647, opened 2024-02-26
Fixes for wasm threads

carlopi:
Various fixes around building of duckdb-wasm and duckdb-wasm compatible loadable extensions in multithreaded setup.


Mytherin:
Thanks!

PR #1651, opened 2024-02-29
Add `allowUnsignedExtensions` to DuckDB Config

Y--:
Right now one need two different builds if they want to switch between a version of DuckDB-WASM that accept unsigned extensions and one that doesn't.
This is not consistent with other platforms (such as CLI which can be started with `-unsigned` or Python which can open a DB with `allow_unsigned_extensions` configuration)

This PR allows to open the DB with this flag as well.


carlopi (approved):
Great, thanks!

Mytherin:
Thanks!

PR #1664, opened 2024-03-11
Use node_modules provided http-server

isaacbrodsky:
`http-server` is installed by yarn but the Makefile assumes it is provided on the path, this changes it to be explicit where to find http-server.


carlopi:
Thanks!

domoritz:
Ups, I never divided my review. Npx would maybe be more robust.

PR #1677, opened 2024-03-18
Bump to DuckDB v0.10.1

carlopi:
This is currently NOT working due to:
* CSV test being too inflexible
* test_all_types test being too inflexible
* duckdb-wasm extensions are not there yet available

1 and 2 will be fixed test side, extensions to be signed/uploaded on a side.


Mytherin:
The CSV test likely fails because of the change in header detection (see https://github.com/duckdb/duckdb/pull/10714).

Mytherin:
Thanks!

Issue #629, opened 2022-02-22
What do COI and EH mean?


zenazn:
Maybe a dumb question, but I wasn't able to find answers on my own by reading the code. What do "COI" and "EH" mean? (I presume "MVP" means "minimum viable product," but maybe that's wrong?)

Alex-Monahan:
I belive coi is [cross origin isolation](https://web.dev/why-coop-coep/), and EH is [exception handling](https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/Exceptions.md)

Does that help?

zenazn:
Yes! Thank you

manzt:
what is MVP?

carlopi:
MVP comes from the WebAssembly jargon, it stands for Minimum Viable Product that is the working-name (that then has been kept around) for the first stable version of the WebAssembly standard.

See for example here: https://github.com/WebAssembly/design/blob/main/MVP.md

manzt:
thank you!!

PR #1685, opened 2024-03-28
Document MVP/EH/COI acroynms

keller-mark:
I had this same question and found https://github.com/duckdb/duckdb-wasm/issues/629


carlopi (approved):
Thanks!

PR #1702, opened 2024-04-15
Support parquet range requests for AWS S3 pre-signed URLs

### LINKED ISSUE #1367: Not getting partial reads of large parquet file in AWS S3 opened from pre-signed URL
### What happens?

I am trying to read a large Parquet file from AWS S3 using pre-signed URLs. The expectation is that DuckDB-WASM will be able to do partial reads of the large file. The observed behavior is the entire file is read. I can reproduce this by first registering the file with `registerFileURL` writing a SELECT against the file alias as well as putting the pre-signed URL inline in a SELECT.

### To Reproduce

1. When using `await this.db.registerFileURL('data.parquet', 'https://the-bucket.s3.amazonaws.com/data.parquet?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=ASI....', DuckDBDataProtocol.HTTP, false);` 

Request 1: HEAD w/ `Range: bytes=0-` header results in 403 w/ `Access-Control-Expose-Headers: Access-Control-Allow-Origin, Accept-Ranges, Content-Range, ETag` header and in the console logs `falling back to full HTTP read for: https://the-bucket.s3.amazonaws.com/data.parquet?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=ASI...`
This is expected since the pre-signed URL only allows GET

Request 2: GET w/ no relevant header results in 200 reading the entire file

2. When using `select column_1 from 'https://the-bucket.s3.amazonaws.com/data.parquet?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=ASI...' limit 1;`

Request 1: HEAD w/ no relevant header results in 403 w/ `Access-Control-Expose-Headers: Access-Control-Allow-Origin, Accept-Ranges, Content-Range, ETag` header
Again, this is expected since the pre-signed URL only allows GET

Request 2: GET w/ `Range: bytes=0-0` header results in 206 w/ `Access-Control-Expose-Headers: Access-Control-Allow-Origin, Accept-Ranges, Content-Range, ETag` header and a single byte `P` returned in the Response body.
Seems to prove `Range` header can be used to fetch a partial file.

But then... instead of starting to partially read the data needed to serve the query, what I observed when using  `registerFileURL` seems to repeat.

Request 3: HEAD w/ `Range: bytes=0-` header results in 403 w/ `Access-Control-Expose-Headers: Access-Control-Allow-Origin, Accept-Ranges, Content-Range, ETag` header
This is expected since the pre-signed URL only allows GET.  But why is it doing a HEAD here, when Request 2 showed the server can do partial reads using GET w/ `Range` header?

Request 4: GET w/ no relevant header results in 200 reading the entire file instead of partial reads.

### Browser/Environment:

Chome115.0.5790.171 (Official Build) (64-bit)

### Device:

Lenovo ThinkPad P16

### DuckDB-Wasm Version:

1.27.0

### DuckDB-Wasm Deployment:

shell.duckdb.org

### Full Name:

mark.lambert@gmail.com

### Affiliation:

Episode-VII Solutions

carstonhernke:
Resolves #1367 

AWS S3 allows for creating pre-signed URLs. However, these pre-signed URLs only allow a single HTTP operation/verb. So a single URL cannot support both HEAD and GET requests.

This causes problems for the current logic which detects whether range requests are supported, because it looks for a successful HEAD request for the Range `bytes=0-`. For an AWS pre-signed URL, this request will fail. However, AWS pre-signed URLs do support range requests.

This PR addresses the issue by adding additional logic which checks whether a GET request with a `Range 0-0` header returns 206. It uses the `Content-Range` response header to determine the remote file size.

 _NOTE: the remote file must have CORS settings that expose the `Content-Range` header._

### DIFF
--- a/packages/duckdb-wasm/src/bindings/runtime_browser.ts
+++ b/packages/duckdb-wasm/src/bindings/runtime_browser.ts
@@ -180,6 +180,30 @@ export const BROWSER_RUNTIME: DuckDBRuntime & {
                             mod.HEAPF64[(result >> 3) + 1] = 0;
                             return result;
                         }
+
+                        // Alternate range request check using GET, if HEAD is not available
+                        const xhrGet = new XMLHttpRequest();
+                        if (file.dataProtocol == DuckDBDataProtocol.S3) {
+                            xhrGet.open('GET', getHTTPUrl(file.s3Config, file.dataUrl!), false);
+                            addS3Headers(xhrGet, file.s3Config, file.dataUrl!, 'GET');
+                        } else {
+                            xhrGet.open('GET', file.dataUrl!, false);
+                        }
+
+                        xhrGet.responseType = 'arraybuffer';
+                        xhrGet.setRequestHeader('Range', `bytes=0-0`);
+                        xhrGet.send(null);
+                        // the content range header is used to determine the length of the entire file
+                        const contentRange = xhrGet.getResponseHeader('Content-Range')?.split('/')[1];
+                        const contentLength2 = xhrGet.getResponseHeader('Content-Length');
+
+                        // supports range requests
+                        if (xhrGet.status == 206 && contentLength2 !== null && +contentLength2 == 1 && contentRange !== undefined) {
+                            const result = mod._malloc(2 * 8);
+                            mod.HEAPF64[(result >> 3) + 0] = +contentRange;
+                            mod.HEAPF64[(result >> 3) + 1] = 0;
+                            return result;
+                        }
                     } catch (e: any) {
                         error = e;
                         console.warn(`HEAD request with range header failed: ${e}`);


carlopi:
Hi @carstonhernke, thanks for the PR, I think this is better unified with the logic within the `if (file.allowFullHttpReads)` bracket, since when we issue a GET request we might still get the whole file back, depending on the server.


Or it can special cased for S3 specifically, possibly via a "HasContentRange()" property that can be false by default but true for S3.

My preference would be for integrating the code you propose with the code below, something like:
```diff
-                        if ((contentLength !== null) && (+contentLength > 1)) {
+                        if (((contentLength !== null) && (+contentLength > 1)) || file.HasContentRange()) {
                            // 2. Send a dummy GET range request querying the first byte of the file
                            //          -> good IFF status is 206 and contentLenght2 is 1
                            //          -> otherwise, iff 200 and contentLenght2 == contentLenght
                            //                 we just downloaded the file, save it and move further
                            const xhr = new XMLHttpRequest();
                            if (file.dataProtocol == DuckDBDataProtocol.S3) {
                                xhr.open('GET', getHTTPUrl(file.s3Config, file.dataUrl!), false);
                                addS3Headers(xhr, file.s3Config, file.dataUrl!, 'GET');
                            } else {
                                xhr.open('GET', file.dataUrl!, false);
                            }
                            xhr.responseType = 'arraybuffer';
                            xhr.setRequestHeader('Range', `bytes=0-0`);
                            xhr.send(null);
                            const contentLength2 = xhr.getResponseHeader('Content-Length');
+                           const contentRange = xhr.getResponseHeader('Content-Range')?.split('/')[1];
-                            if (xhr.status == 206 && contentLength2 !== null && +contentLength2 == 1) {
+                           if (xhr.status == 206 && contentLength2 !== null && +contentLength2 == 1 && contentLength !== null) {
                                const result = mod._malloc(2 * 8);
                                mod.HEAPF64[(result >> 3) + 0] = +contentLength;
                                mod.HEAPF64[(result >> 3) + 1] = 0;
                                return result;
                            }
+.                         // This can also be before, unsure who should get precedence
+.                         // Actually this has likely to be before
+                           if (xhr.status == 206 && contentRange !== null) {
+                               const result = mod._malloc(2 * 8);
+                               mod.HEAPF64[(result >> 3) + 0] = +contentRange;
+                               mod.HEAPF64[(result >> 3) + 1] = 0;
+                               return result;
+                           }
                            if (xhr.status == 200 && contentLength2 !== null && +contentLength2 == +contentLength) {
                                console.warn(`fall back to full HTTP read for: ${file.dataUrl}`);
                                const data = mod._malloc(xhr.response.byteLength);
                                const src = new Uint8Array(xhr.response, 0, xhr.response.byteLength);
                                mod.HEAPU8.set(src, data);
                                const result = mod._malloc(2 * 8);
                                mod.HEAPF64[(result >> 3) + 0] = xhr.response.byteLength;
                                mod.HEAPF64[(result >> 3) + 1] = data;
                                return result;
                            }
                        }
```

mostly to avoid throwing all informations away when we happen to have just downloaded the whole file (that given we don't know it might as well happen) and to keep the logic together / more well tested.

carstonhernke:
Sounds good. I will leave some comments in https://github.com/duckdb/duckdb-wasm/pull/1717

PR #1727, opened 2024-05-04
Refactor react-duckdb

willium:
The existing `react-duckdb` providers were pretty convoluted and had properties that it didn't use. I refactored it and added documentation and comments.


ncsmikewoods:
I've been keeping an eye on these changes.  It looks like they'll address some problems with race conditions in duckdb initialization.  I've been holding off rewriting my connection management code because this refactor looks like a much cleaner implementation that's what's there.

Thanks for your work @willium

jr200:
This patch would be helpful to me too. Are there plans to merge and release this PR? Thanks!

carlopi:
I have to say I don't feel very competent to review this PR, would someone volunteer to have a deeper look?

packages/react-duckdb/src/duckdb_provider.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+type DuckDBContextValue = {
+    database: AsyncDuckDB | null;
+    connectionPool: ConnectionPool;
+    establishConnection: (id?: number) => Promise<void>;
+};
+
+const DuckDBContext = createContext<DuckDBContextValue | null>(null);
+
+// TODO: consider adding support for passing in an existing AsyncDuckDB instance
```
qw-in, line 40: My $0.02 is it seems like most of the complexity is to handle the `bundle`, `bundles`, etc

I think this is a wise comment, and I even wonder if supporting _only_ an existing `AsyncDuckDB` instance makes sense

qw-in (commented):
Looks like `index.ts` is still referencing the deleted files here

https://github.com/duckdb/duckdb-wasm/blob/cda56b44b3d62e2f30e423a9c4c027adf114fb7a/packages/react-duckdb/src/index.ts#L1-L4

PR #1768, opened 2024-06-17
Handle longer queries

carlopi:
duckdb_web_query_run implicitly relied on stack-allocated strings for handling `ccall` Emscripten mechanisms.

Move to explicit allocations in the HEAPU8 buffer and passing around only pointers.

Before this fix queries at around 64K would become buggy, now up to 1e7, possibly even more can be handled.

This fixes a long standing issue in https://github.com/duckdb/duckdb-wasm/issues/1477


carlopi:
@antonycourtney to be credited with the parametrisable test, thanks.

antonycourtney:
Awesome that you tracked this down, @carlopi !  Can't wait to get this fix into MotherDuck!

Issue #1696, opened 2024-04-09
INTERVAL bug (wrong endian?)


david542542:
### What happens?

Running `SELECT INTERVAL '2 year'` (or really any other `INTERVAL`) produces the wrong output in the wasm shell. It seems like  it has to do with endianness: 2 years is 24 months and here 24 is put  into fractional seconds, i.e at the other end of the number.

<img width="536" alt="image" src="https://github.com/duckdb/duckdb/assets/909771/b483b7f0-1aa7-4af3-9fb1-e073773c00e4">

### To Reproduce

You can run this query [here](https://shell.duckdb.org/#queries=v0,select-interval-'2-year'~) (click to open in wasm shell). 

`SELECT INTERVAL '2 day'`

### OS:

Mac > Chrome > DuckDB Wasm Shell

### DuckDB Version:

v0.10.1

### DuckDB Client:

Wasm

### Full Name:

David Litwin

### Affiliation:

None

### Have you tried this on the latest [nightly build](https://duckdb.org/docs/installation/?version=main)?

I have tested with a nightly build

### Have you tried the steps to reproduce? Do they include all relevant data and configuration? Does the issue you report still appear there?

- [X] Yes, I have

Mause:
@Mytherin I think this is actually an issue with DuckDB core's arrow layer, as I see the same issue with duckdb-rs. I did briefly discuss this with @Maxxen as well

Mytherin:
Why does the problem then not occur in Python?

```py
>>> import duckdb
>>> duckdb.sql("SELECT INTERVAL '2 year'").arrow()
# pyarrow.Table
# CAST('2 year' AS INTERVAL): month_day_nano_interval
# ----
# CAST('2 year' AS INTERVAL): [[24M0d0ns]]

```

Mytherin:
Actually since both `duckdb-rs` and `duckdb-wasm` use the Rust arrow library, is it perhaps possible the problem is in the Rust Arrow library?

Mause:
That's possible yeah, I'd assumed that wasm used arrow-js, but it does seem to use arrow-rs yes

Max commented in Discord:
>  Sounds like a lifetime issue with the arrow return values then, might be some arrow array destructor we set too early or something. Will have a look

Mause:
Looks like they did have issues here at one point too: https://github.com/apache/arrow-rs/pull/2235/files

Mytherin:
[...]

Issue #448, opened 2021-12-07
histogram causes crash


stratospark:
The following currently breaks on https://shell.duckdb.org/. It works fine on my CLI.

```
DuckDB Web Shell
Database: v0.3.1-dev535
Package:  @duckdb/duckdb-wasm@1.11.1-dev64.0

Connected to a transient in-memory database.
Enter .help for usage hints.

duckdb> create table t1 (i integer);
┌┐
└┘

duckdb> insert into t1 values (1),(2),(1);
┌───────┐
│ Count │
╞═══════╡
│     3 │
└───────┘

duckdb> select * from t1;
┌───┐
│ i │
╞═══╡
│ 1 │
│ 2 │
│ 1 │
└───┘

duckdb> select histogram(i) from t1;
```

error in browser console:

```
f272ce859168ca188950.wasm:0x1b25e3 Uncaught (in promise) RuntimeError: unreachable
    at f272ce859168ca188950.wasm:0x1b25e3
    at f272ce859168ca188950.wasm:0x1b8514
    at f272ce859168ca188950.wasm:0x1bc638
    at f272ce859168ca188950.wasm:0x1bc0b6
    at f272ce859168ca188950.wasm:0x1bc66e
    at f272ce859168ca188950.wasm:0xda902
    at f272ce859168ca188950.wasm:0x1411a2
    at f272ce859168ca188950.wasm:0x1402ea
    at f272ce859168ca188950.wasm:0xcc214
    at f272ce859168ca188950.wasm:0x1bd805
```

expected result
```
D .version
SQLite v0.3.1 88aa81c6b
msvc-1929

D select histogram(i) from t1;
┌──────────────┐
│ histogram(i) │
├──────────────┤
│ {1=2, 2=1}   │
└──────────────┘
```

ankoh:
Thanks for reporting this. That's a bug in the shell.
DuckDB-Wasm itself will return an arrow.Map_.

https://github.com/duckdb/duckdb-wasm/blob/ffa9210415e394fa84f0c769a54ef86b169bc2ab/packages/duckdb-wasm/test/regression/github_448.test.ts#L18-L33

sgilligan:
@ankoh will the shell display the result? When I try the shell shows an elapsed time count by nothing else

Issue #1133, opened 2023-01-27
SELECT INTERVAL 1 DAY hangs the online shell


orlp:
To reproduce execute the following on https://shell.duckdb.org/:

```sql
SELECT INTERVAL 1 DAY;
```

The error in the Javascript console is

```
125.6dc9a38060e9fe5bb98b.js:1 panicked at 'already mutably borrowed: BorrowError', packages/duckdb-wasm-shell/crate/src/shell.rs:1064:29
```

Unfortunately the Rust stacktrace is empty.

carlopi:
Update: now it's not trapping in the front end anymore!
(but conversions from internal interval representation to Arrow representation needs to be worked on, since here it currently returns:
`0 years 0 mons 0 days 0 hours 0 mins 4.294967296 secs`

carlopi:
And this seems to be the same problem as #1203.

carlopi:
Thanks a lot, this is now solved thanks to https://github.com/duckdb/duckdb-wasm/pull/1769

PR #1769, opened 2024-06-17
Bump arrow to 16.0.0

carlopi:
Bump arrow version to Arrow-JS 16 or Arrow-Rust 52.

This allows to fix a few open bugs:
* fixes https://github.com/duckdb/duckdb-wasm/issues/1696, where the interval definition was using the wrong indian representation. Thanks @Maxxen, @Mytherin, @david542542, @pitrou and others for the contributions.
* fixes https://github.com/duckdb/duckdb-wasm/issues/448, where histogram would create a map type, that was not handled at the arrow-rust level used in the shell.
* fixes (similar to above) representations for `FROM duckdb_functions()`, that since 1.0.0 have a map field

I need to review other open issues, there might be other somehow connected to this.

Big thanks in general to the Arrow's maintainers.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,20 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "ahash"
+version = "0.8.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


carlopi:
More interval bugs connected to bumping arrow-rust:
fixes https://github.com/duckdb/duckdb-wasm/issues/1203
fixes https://github.com/duckdb/duckdb-wasm/issues/1204
fixes https://github.com/duckdb/duckdb-wasm/issues/1133

carlopi:
I would merge this, since this solves quite some problems and current limitations, but a deeper look from @domoritz would be very welcome.

PR #1803, opened 2024-07-31
Fix COI compilation

carlopi:
#### Problem: extern C and includes are not friends
Arrow submodule vendors in [xxhash](https://github.com/Cyan4973/xxHash), and the vendored in apache-arrow does a funny:
```
extern 'C' {

#include <something>

} 
```

Problem? This propagates C-linkage to the header, that then will eventually `#include <emscripten.h>`, that since it relies on  template code can't be compiled with C-linkage.

[xxhash](https://github.com/Cyan4973/xxHash) should move includes outside of `extern C` brackets, I will raise a bug report there.

I am not sure/I don't see if situation can be detected / handled at the emscripten.h level. Basic open question is: can a header know it's being compiled within a `extern 'C'` namespace?

#### Solution: ... patches!
As always, adding more layers of patching solves (most) problems connected to customising submodules, at least in the short term. Here the patch explicitly includes emscripten.h outside of the extern C code.

I moved other patches from `duckdb_patches` to the more structured `patches/duckdb`, so that now also a `patches/arrow` can be added.


carlopi:
Issue raising this on the Cyan4973/xxHash repository: https://github.com/Cyan4973/xxHash/issues/960.

For the patch to be removed that has to go in, a new version being released, and arrow bumping vendored dependency to newer version, and us bumping arrow dependency.

Issue #1888, opened 2024-10-09
Failed to build for duckdb_wasm sub-project test (tester)


ritalin:
### What happens?

I've challenged to run `duckdb_wasm` sub-project test (tester).

Following `README`, I've tried to build but failed.

```
$ cmake --build .
...
Undefined symbols for architecture x86_64:
  "duckdb_skiplistlib::skip_list::_throw_exceeds_size(unsigned long)", referenced from:
```

This is because of not specifying `libduckdb_skiplistlib.a` to a cmake configuration (`lib/cmake/duckdb.cmake`).

### To Reproduce

```
mkdir -p build
cd build
cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Debug ../lib
cmake --build .
```

### Browser/Environment:

MacOS Ventura 13.6.9 / Apple clang version 14.0.3

### Device:

Macbook Air 2020 (Intel Core i7 1.2GHz)

### DuckDB-Wasm Version:

1.29.0 (currently main stream)

### DuckDB-Wasm Deployment:

None (because of unittest problem)

### Full Name:

Kazuhiko TAMURA

### Affiliation:

have no job

ritalin:
Putting `libduckdb_skiplistlib.a` dependency to `target_link_libraries` of `lib/cmake/duckdb.cmake` led to resolve this build probjem.


```
target_link_libraries(
  duckdb
  INTERFACE ${install_dir}/lib/libduckdb_re2.a
  ...
  INTERFACE ${install_dir}/lib/libduckdb_skiplistlib.a
  ...)
```

ritalin:
> But many test cases has failed...
14/102 test cases have failed.

Following

* AllTypesTest.FullRangeTypes
* FilePageBufferTest.FixSingle
* FilePageBufferTest.FIFOEviction
* FilePageBufferTest.LRUEviction
* FilePageBufferTest.ParallelFix
* FilePageBufferTest.ParallelExclusiveAccess
* FilePageBufferTest.ParallelScans
* FilePageBufferTest.ParallelReaderWriter
* CSVExportTest.TestExport
* ParquetLoadTest.LoadParquet
* ParquetLoadTest.LoadParquetTwice
* FileSystemBufferTest.FlushFrameMemoryBugRegression
* WebFileSystemTest.LoadParquet

ritalin:
I'm ready for PR for this issue.

carlopi:
A PR is very welcome

[...]

PR #1890, opened 2024-10-10
Fix to build failure for duckdb_wasm sub-project test

ritalin:
### Description

This is PR for issue #1888.
A test for `duckdb_wasm` sub-project has failed to build.
It resolves this issue.

### Out-of-scope

To build test artifact is success, but executing test still fails.
Following testcases:

#### AllTypesTest.FullRangeTypes

This test is used a udf `test_all_types`.
It fails to not match expects columns with `test_all_types` columns.
It will be easy modifying.

### ParquetLoadTest.LoadParquet

In 1.29.0, DuckDB extensions support enabled by default.
So calling `duckdb_web_parquet_init(&db->database())` failes.
It will be easy modifying. (line-delete only)

Same for the following

* ParquetLoadTest.LoadParquetTwice
* FileSystemBufferTest.FlushFrameMemoryBugRegression
* WebFileSystemTest.LoadParquet

### CSVExportTest.TestExport

This test fails calling `WebDB::Connection::InsertCSVFromPath`.

```
{"exception_type":"Not implemented","exception_message":"Unsupported compression type for default file system"}
```

But this method does not have a compression option.
It makes no sense for me.

### FilePageBufferTest.FixSingle

It fails calling `TestableFilePageBuffer::BuffersFile` after `FilePageBuffer::FileRef::Release`.
It seems that accessing released file buffer is incorrect.
It will need to make this specification clear.

### DIFF
--- a/lib/cmake/duckdb.cmake
+++ b/lib/cmake/duckdb.cmake
@@ -87,6 +87,7 @@ target_link_libraries(
   INTERFACE ${install_dir}/lib/libduckdb_miniz.a
   INTERFACE ${install_dir}/lib/libduckdb_mbedtls.a
   INTERFACE ${install_dir}/lib/libduckdb_yyjson.a
+  INTERFACE ${install_dir}/lib/libduckdb_skiplistlib.a
   INTERFACE ${install_dir}/lib/libduckdb_pg_query.a
   INTERFACE ${install_dir}/lib/libduckdb_utf8proc.a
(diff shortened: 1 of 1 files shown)


ritalin:
I can include the fixes `AllTypesTest.FullRangeTypes` and `ParquetLoadTest.LoadParquet` this PR. 
What do you think?

Issue #447, opened 2021-12-06
Support array arguments in prepared statements


domoritz:
See https://github.com/duckdb/duckdb-wasm/issues/441#issuecomment-986913723

ankoh:
Right now, we're just `JSON.stringify` the arguments here:

https://github.com/duckdb/duckdb-wasm/blob/ffa9210415e394fa84f0c769a54ef86b169bc2ab/packages/duckdb-wasm/src/bindings/bindings_base.ts#L197-L210

and then unpack them here:

https://github.com/duckdb/duckdb-wasm/blob/ffa9210415e394fa84f0c769a54ef86b169bc2ab/lib/src/webdb.cc#L219-L242

We probably just need to extend rapidjson::Document -> std::vec<duckdb::Value>.

scamden:
any movement here? would love to use this

scamden:
in my case i hacked around this by doing:
```
const bindingIndexes = bartRowIds.map((_, i) => `?`).join(',');
   `SELECT * FROM table WHERE row_id in (${bindingIndexes})`,
              { bindings: bartRowIds }
```

but a little tedious and would be especially tricky if i had any other bindings

mbostock:
Apologies for bumping, but I would also love to see this fixed — we are now using DuckDB-Wasm for SQL code cells in Observable Framework.

ritalin:
@ankoh
Is this work in progress?
I have a plan modifying PR.

PR #1892, opened 2024-10-11
Supports array arguments in prepared statements

ritalin:
## Description

This PR implements in supporting an array arguments in the prepared statements (issue #447).

## Usage

You can pass Javascript's array as argument for `query` / `send` method of `AsyncPreparedStatement`.

```js
const stmt = await conn.prepare(`SELECT * FROM range(1, ?::int) t(id) WHERE id = ANY(?)`);
const result = await stmt.query(100, [2, 3, 5, 7, 11]); 
```

`query` / `send` method of `PreparedStatement` can also apply similarly.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -361,6 +361,7 @@ if(NOT EMSCRIPTEN)
       ${CMAKE_SOURCE_DIR}/test/json_dataview_test.cc
       ${CMAKE_SOURCE_DIR}/test/memory_filesystem_test.cc
       ${CMAKE_SOURCE_DIR}/test/parquet_test.cc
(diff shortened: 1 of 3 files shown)


Fil:
I suppose you mean

```diff
- SELECT * FROM range(1, ?::int) t(id) WHERE id = IN(SELECT unnest(?))
+ SELECT * FROM range(1, ?::int) t(id) WHERE id = ANY(SELECT unnest(?))
```

ritalin:
@Fil 
This is type

The following is correct:
```sql
SELECT * FROM range(1, ?::int) t(id) WHERE id IN(SELECT unnest(?));
```

Sorry.

PR #1893, opened 2024-10-11
Bump DuckDB to f680b7d08f (v1.1.2)

carlopi:


carlopi:
Workflow is expected to fail due to missing extensions. I am having them built / deployed, then I will re-run here.

PR #1903, opened 2024-10-28
Add `allow_stream_result` to `startPendingQuery`

Y--:
This PR adds a new parameter to `startPendingQuery` API to be forwarded to DuckDB's [PendingQuery](https://github.com/duckdb/duckdb/blob/cfc08c1b336fa74356df888d02d752c354887c01/src/include/duckdb/main/client_context.hpp#L112)

Notes:
* parameter is optional
* default value is `false` to be consistent with [Connection](https://github.com/duckdb/duckdb/blob/cfc08c1b336fa74356df888d02d752c354887c01/src/include/duckdb/main/connection.hpp#L96) and existing behavior


carlopi (approved):
Thanks a lot!!

PR #1905, opened 2024-10-30
Fix string passing at interface C++ to JS

carlopi:
Thanks to @jraymakers that identified the problem.

There are still a few places where we pass table names or DB names, those ought to be shorter, but I will address and remove those in a separate PR, for now keeping old and new interface in, but using only the new one where we pass only malloc-ed memory around.

### DIFF
--- a/packages/duckdb-wasm/src/bindings/bindings_base.ts
+++ b/packages/duckdb-wasm/src/bindings/bindings_base.ts
@@ -139,12 +139,11 @@ export abstract class DuckDBBindingsBase implements DuckDBBindings {
         const bufferOfs = this.mod.HEAPU8.subarray(bufferPtr, bufferPtr + BUF.length );
         bufferOfs.set(BUF);
         const [s, d, n] = callSRet(this.mod, 'duckdb_web_tokenize_buffer', ['number', 'number'], [bufferPtr, BUF.length]);
+        this.mod._free(bufferPtr);
         if (s !== StatusCode.SUCCESS) {
-            this.mod._free(bufferPtr);
             throw new Error(readString(this.mod, d, n));
         }
         const res = readString(this.mod, d, n);
-        this.mod._free(bufferPtr);
         dropResponseBuffers(this.mod);
         return JSON.parse(res) as ScriptTokens;
     }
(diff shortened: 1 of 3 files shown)


packages/duckdb-wasm/src/bindings/bindings_base.ts
```diff
@@ -134,11 +134,17 @@ export abstract class DuckDBBindingsBase implements DuckDBBindings {
...
     /** Tokenize a script */
     public tokenize(text: string): ScriptTokens {
-        const [s, d, n] = callSRet(this.mod, 'duckdb_web_tokenize', ['string'], [text]);
+        const BUF = TEXT_ENCODER.encode(text);
+        const bufferPtr = this.mod._malloc(BUF.length );
+        const bufferOfs = this.mod.HEAPU8.subarray(bufferPtr, bufferPtr + BUF.length );
+        bufferOfs.set(BUF);
+        const [s, d, n] = callSRet(this.mod, 'duckdb_web_tokenize_buffer', ['number', 'number'], [bufferPtr, BUF.length]);
         if (s !== StatusCode.SUCCESS) {
+            this.mod._free(bufferPtr);
```
Y--, line 143: Shouldn't it be freed after reading the string? (same below in the 2 other error cases?)

carlopi: I think it's a valid question, but interface should be that duckdb_web_tokenize_buffer should only get a view of the C-array (ex-string), but are not allowed to actually reference to it and need to do a copy if query needs to be reported back in the error.

But then if this is the case, maybe code is more clear with the free just after the call. Let me double check this.

jraymakers:
Should `createPrepared` get the same treatment? It also takes a query string.

packages/duckdb-wasm/src/bindings/bindings_base.ts
carlopi: I think I cleaned up a bit by moving the free up

carlopi:
> Should `createPrepared` get the same treatment? It also takes a query string.

Yes, thanks. I will address those in a separate PR, I think this still makes sense as-is.

jraymakers:
> > Should `createPrepared` get the same treatment? It also takes a query string.
> 
> Yes, thanks. I will address those in a separate PR, I think this still makes sense as-is.

Separate PR is fine. But note that `createPrepared` is also an important call for us.

carlopi:
I am onto that.

jraymakers:
This is minor, but a couple of the C functions have unused "script" arguments:
- `duckdb_web_pending_query_poll`
- `duckdb_web_pending_query_cancel`

carlopi:
Follow up is here: https://github.com/duckdb/duckdb-wasm/pull/1906

PR #1913, opened 2024-11-10
Export to JS also invoke_ functions

carlopi:
This is needed for Rust-based extensions, that otherwise when dynamically linked will miss `invoke_iii` & family functions.


carlopi:
This is an hack to work around the problem at: https://github.com/emscripten-core/emscripten/issues/22906 and allow to move further in supporting Rust extensions in Wasm.

PR #1918, opened 2024-11-12
Add duckdb patch signature_capi

carlopi:
Signature of capi extensions changed from returning void to returning a boolean, mismatch on native targets is fine, since result is just ignored, but flagged problems in wasm validation due to signature changing in an incompatible way from `(int,int)->()` to `(int,int)->(int)`.

Given the error is on validation of indirect call, I am not aware of a workaround Wasm side, and given capi extensions are still experimental, and more so in duckdb-wasm, I would propose taking this liberty from DuckDB source code to allow more simpler exploration on this matter.

Discussed with @samansmink.

### DIFF
--- a/patches/duckdb/signature_capi.patch
+++ b/patches/duckdb/signature_capi.patch
@@ -1,0 +1,13 @@
+diff --git a/src/main/extension/extension_load.cpp b/src/main/extension/extension_load.cpp
+index 59fc4e8cd9..b0282a7103 100644
+--- a/src/main/extension/extension_load.cpp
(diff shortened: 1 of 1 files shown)


samansmink (approved):

carlopi:
Also to be noted, this changed in duckdb main branch already, so that would allow to test extensions either:
* duckdb native on v1.1.3
* duckdb native on main
* duckdb-wasm on v1.1.3
* duckdb-wasm on main

and see similar results. Otherwise the third option would have been left behind, and that would mean that errors will be only noticed with delays.

PR #1925, opened 2024-11-26
Fix return type of `AsyncDuckDBConnection.prepare`

swlynch99:
This looks like it is supposed to return `AsyncPreparedStatement<T>` but since the type parameter is not provided it actually returns `AsyncPreparedStatement<any>`. That passes type checking anyway since `any` is compatible with `T` but is inconvenient downstream since it doesn't actually use the type parameter.


carlopi (approved):
Welcome!
This makes sense, thanks for catching this and sending the PR.

Thanks!!

PR #1927, opened 2024-12-09
[Experimental] Tracking duckdb main branch

carlopi:
This tracks current main in duckdb/duckdb (1c6ea28bdc7557dddd3726bb3c68a1fcc7a4ede1), and will do so in the run-up to duckdb v1.2.0.

Signed extensions are not available yet for that hash.


carlopi:
This is currently tracked in the https://github.com/duckdb/duckdb-wasm/tree/refs/heads/v1.2-histrionicus branch

PR #1933, opened 2025-01-07
Update core extensions URL

mattbasta:
extensions.duckdb.org no longer exists; the extensions are curated in the docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Supported DuckDB features:
 
 DuckDB is extensible and this allows to delegate functionality to [extensions](https://duckdb.org/docs/extensions/overview).
 
-Core extensions are available at https://extensions.duckdb.org, and community extensions are available at https://community-extensions.duckdb.org.
+Core extensions are available at https://duckdb.org/docs/extensions/core_extensions, and community extensions are available at https://community-extensions.duckdb.org.
 ```sql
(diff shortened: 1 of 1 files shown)


carlopi:
Hi, thanks for the contribution, I will revise the text but what's trying to convey is that the core extension endpoint is `https://extensions.duckdb.org` (even though extensions docs are elsewhere).

We could / should probably set up redirects, since those might help.

carlopi:
@mattbasta: you are very welcome to send a PR that reword this / adds some context, but I think just changing the link I don't think cleans it up.

mattbasta:
Hey @carlopi. The URL that you mentioned in your comment isn't a working link. Perhaps I'm not understanding what you mean by "endpoint" here; is the purpose of the text not to direct folks to be able to see the extensions?

I do agree that getting extensions.duckdb.org to redirect would be preferable, though.

carlopi:
I guess the list of core extensions is at https://duckdb.org/docs/extensions/core_extensions.

What do I meant is that if doing `INSTALL spatial;` there would be a network request to 'https://community-extensions.duckdb.org/v1.1.3/wasm_eh/spatial.duckdb_extension.wasm'.

Thanks to this PR I see this not being clear at all.

PR #1941, opened 2025-01-22
Query progress updates to be passed to main thread

carlopi:
Main note is that this allows while busy executing (and so no messages can be received by the worker) messages can still be sent to the main/UI thread.

Missing:
- [ ] Add a demo/test

### DIFF
--- a/lib/include/duckdb/web/webdb.h
+++ b/lib/include/duckdb/web/webdb.h
@@ -33,6 +33,8 @@ class WebDB {
    public:
     /// A connection
     class Connection {
+        friend WebDB;
+
        protected:
         /// The webdb
         WebDB& webdb_;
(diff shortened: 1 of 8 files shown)


carlopi:
@domoritz: thanks for taking a look, nice observation.

I would think that here a 1-way only message via the callback is OK here, and should allow this to be driven by actual progress, but I think I might be missing something, could you expand on how do you see this working?

carlopi:
I would be for merging this, this is still experimental, can always be fiddle with, but I would be interested in your idea.

domoritz:
A readable stream could update the progress. It could look like the random number stream in https://dev.to/ruben_alapont/stream-your-way-creating-custom-streams-395o. A stream is very similar to a js iterator but it can push updates rather than the consumer pulling/iterating.

My question is mainly about API. Callbacks are somewhat older way of doing this in js and these days async, promises, streams, iterators, etc are the way to go.

PR #1960, opened 2025-02-19
fix(bindings): handle accessMode in runtime bindings with OPFS

sumimakito:
> This is an amazing project with many great works. Thank you

# Context

During my recent usage with `duckdb-wasm` with the recent OPFS support done by @e1arikawa (Thank you for the cool work!), I encountered an issue where the file/directory handles were always created even when I opened the database with read-only options like:

```js
db.open({
    path: 'opfs://not_yet_existing.db',
    accessMode: DuckDBAccessMode.READ_ONLY, // We should NOT create any files/directories
})
```

# Summary

Going deeper, it seems we didn't handle the `accessMode` in helper functions like `prepareFileHandles`. Therefore, in this pull request, I tried to add the missing parameters to these functions as optional ones and added several test cases for open operations as such with different `accessMode` in OPFS.

### DIFF
--- a/packages/duckdb-wasm/src/bindings/runtime_browser.ts
+++ b/packages/duckdb-wasm/src/bindings/runtime_browser.ts
@@ -4,6 +4,7 @@ import {addS3Headers, getHTTPUrl} from '../utils';
 import {
     callSRet,
     dropResponseBuffers,
+    DuckDBAccessMode,
     DuckDBDataProtocol,
     DuckDBFileInfo,
     DuckDBGlobalFileInfo,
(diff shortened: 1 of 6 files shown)


packages/duckdb-wasm/src/bindings/runtime_browser.ts
```diff
@@ -133,15 +137,21 @@ export const BROWSER_RUNTIME: DuckDBRuntime & {
...
                 }
                 const fileHandle = await dirHandle.getFileHandle(fileName, { create: false }).catch(e => {
                     if (e?.name === 'NotFoundError') {
-                        console.debug(`File ${path} does not exists yet, creating...`);
-                        return dirHandle.getFileHandle(fileName, { create: true });
+                        if (!isReadOnly) {
+                            console.debug(`File ${path} does not exists yet, creating...`);
+                            return dirHandle.getFileHandle(fileName, { create: true });
+                        }
+                        console.debug(`File ${path} does not exists, aborting as we are in read-only mode`);
```
amiller-gh, line 154: Five lines below this, where we call `createSyncAccessHandle`, we can pass a new optional `mode` to OPFS to further enforce the access mode: https://developer.mozilla.org/en-US/docs/Web/API/FileSystemFileHandle/createSyncAccessHandle#mode

If the access mode is passed to `createSyncAccessHandle`, it would allow read-only connections to be opened simultaneously (for example when using the same app in multiple tabs). A big win for duckdb-wasm + OPFS.

The option is only implemented in Chrome and Edge right now, but issues exist to implement in both Webkit and Firefox:
https://bugs.webkit.org/show_bug.cgi?id=283959
https://bugzilla.mozilla.org/show_bug.cgi?id=1949462

And tests have already been added to the Web Platform Tests dashboard:
https://wpt.fyi/results/fs/FileSystemFileHandle-sync-access-handle-lock-modes.https.tentative.worker.html?label=experimental&label=master&aligned

amiller-gh: Ah, just noticed you started a new PR for this! Will comment over there instead.

sumimakito: Thank you very much for the detailed suggestion. Let me close this PR and try to address that in #1962.

sumimakito:
Moving to #1962

PR #1962, opened 2025-02-22
fix: respect access mode and file open flags

sumimakito:
Previous PR: https://github.com/duckdb/duckdb-wasm/pull/1960 (it can be closed if this PR looks better as this PR addresses more issues for both OPFS and Node FS)

# Summary

* `accessMode` in `db.open(...)` will be respected and empty database files (and directories in OPFS) will no longer be created in the read-only mode
* Fix file open flag used in Node FS and
* Fix the return value of `openFile` in the Node runtime as it should return `nullptr` when the file does not exist and the `FILE_FLAGS_NULL_IF_NOT_EXISTS` is set
* Add missing flags in `FileFlags` to the JavaScript bindings
* Add tests for Node FS and OPFS
* Update patch for `src/storage/storage_manager.cpp` in the `duckdb` submodule

### DIFF
--- a/packages/duckdb-wasm/test/nodefs.test.ts
+++ b/packages/duckdb-wasm/test/nodefs.test.ts
@@ -1,8 +1,14 @@
 import * as duckdb from '../src/';
-import { tmpdir } from 'os';
 import { randomUUID } from 'crypto';
-import path from 'path';
 import { unlink } from 'fs/promises';
+import { mkdirSync } from 'fs';
+import path from 'path';
+
+const tmpdir = () => {
+    const tmp = path.resolve(__dirname, '../test_temp');
+    mkdirSync(tmp, { recursive: true });
+    return tmp;
+}
 
 export function testNodeFS(db: () => duckdb.AsyncDuckDB): void {
     const files: string[] = [];
--- a/packages/duckdb-wasm/.gitignore
+++ b/packages/duckdb-wasm/.gitignore
@@ -6,4 +6,5 @@
 /duckdb_webapi.*
 /dist
 /docs
+/test_temp
 *.d.ts.map


packages/duckdb-wasm/src/bindings/runtime_browser.ts
```diff
@@ -135,13 +139,16 @@ export const BROWSER_RUNTIME: DuckDBRuntime & {
...
                 }
                 const fileHandle = await dirHandle.getFileHandle(fileName, { create: false }).catch(e => {
                     if (e?.name === 'NotFoundError') {
-                        console.debug(`File ${path} does not exists yet, creating...`);
-                        return dirHandle.getFileHandle(fileName, { create: true });
+                        if (isReadWrite) {
+                            console.debug(`File ${path} does not exists yet, creating...`);
+                            return dirHandle.getFileHandle(fileName, { create: true });
+                        }
+                        console.debug(`File ${path} does not exists, aborting as we are in read-only mode`);
```
amiller-gh, line 151: Five lines below this, where we call createSyncAccessHandle, we can pass a new optional mode to OPFS to further enforce the access mode: https://developer.mozilla.org/en-US/docs/Web/API/FileSystemFileHandle/createSyncAccessHandle#mode

If the access mode is passed to createSyncAccessHandle, it would allow read-only connections to be opened simultaneously (for example when using the same app in multiple tabs). A big win for duckdb-wasm + OPFS.

The option is only implemented in Chrome and Edge right now, but issues exist to implement in both Webkit and Firefox:
https://bugs.webkit.org/show_bug.cgi?id=283959
https://bugzilla.mozilla.org/show_bug.cgi?id=1949462

And tests have already been added to the Web Platform Tests dashboard:
https://wpt.fyi/results/fs/FileSystemFileHandle-sync-access-handle-lock-modes.https.tentative.worker.html?label=experimental&label=master&aligned

e1arikawa:
@sumimakito 
I believe this PR should be considered complete without handling separate window or tab processes.

packages/duckdb-wasm/src/bindings/runtime_browser.ts
e1arikawa: @amiller-gh 
I plan to handle separate windows and tabs in a separate PR, so I believe it's fine for this PR to focus only on a single window.

carlopi:
Thanks a lot for the PR, I will test and send a review in the next hours.

packages/duckdb-wasm/test/nodefs.test.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    afterAll(async () => {
+        await Promise.all(files.map(file => unlink(file).catch(() => {})));
+        await db().flushFiles();
+        await db().dropFiles();
+    });
+
+    describe('Node FS', () => {
+        it('Should not create an empty DB file in read-only mode for non-existent path', async () => {
+            const tmp = tmpdir();
```
carlopi, line 18: Usage of `tmpdir()` seems to not work in the CI test setup, could you have a look at the failure?

sumimakito: That's okay. I was also trying to run tests in my local environment with some of the intermediate artifacts produced by the CI to reduce the build time, but it didn't run smoothly as the CI did 😿

M-Taan:
Hey, amazing work thank you for this @sumimakito
Just checking on the status of this PR and if there's anything I could help with to get it merged asap?

PR #1977, opened 2025-03-12
Added a Check for the Worker is Running or Not

### LINKED ISSUE #1970: send function make infinite loop on destroyed worker
### What happens?

In the send function, when the worker is destroyed, there is a while loop which fail indefinitely, making the website failling completely

Destroy can occur any time and the send function can be called in a promise later.

```
// Infinite error like this
hook.js:608 cannot send a message since the worker is not set!
overrideMethod      @hook.js:608
postTask            @async_bindings.ts:103
pollPendingQuery    @async_bindings.ts:422
send                @async_connection.ts:65
```

### To Reproduce

There is a while loop in the send function which doesnt stop when there is an error

https://github.com/duckdb/duckdb-wasm/blob/main/packages/duckdb-wasm/src/bindings/connection.ts#L43
```
while (header == null) {
    header = await new Promise((resolve, reject) => {
        try {
            resolve(this._bindings.pollPendingQuery(this._conn));
        } catch (e: any) {
            console.log(e);
            reject(e);
        }
    });
 }
```

https://github.com/duckdb/duckdb-wasm/blob/main/packages/duckdb-wasm/src/parallel/async_connection.ts#L64
```
while (header == null) {
    header = await this._bindings.pollPendingQuery(this._conn);
}
```

Destroy a worker and call send function

```
const c = await db.connect();
await db.terminate(); // Usually called somewhere else
await c.send('SELECT 1;');
await c.close();
```

### Browser/Environment:

Chrome 133

### Device:

Macbook

### DuckDB-Wasm Version:

1.29.1-dev17.0

### DuckDB-Wasm Deployment:

https://shell.duckdb.org/

### Full Name:

Fontaine Audrien

### Affiliation:

Unify

parthbagda211:
Fixes #1970

This pull request includes an important change to the error handling in the `DuckDBConnection` class within the `connection.ts` file. The change improves the handling of specific error messages related to worker termination.

Error handling improvement:

* [`packages/duckdb-wasm/src/bindings/connection.ts`](diffhunk://#diff-34de9050bae040fbfa05ede411b102748cc85f40311763fe461a0a6446208e49R49-R57): Added a check for the error message "worker is not set!" to provide a more specific error message indicating that the worker has been terminated.

### DIFF
--- a/packages/duckdb-wasm/src/bindings/connection.ts
+++ b/packages/duckdb-wasm/src/bindings/connection.ts
@@ -46,7 +46,14 @@ export class DuckDBConnection {
                     resolve(this._bindings.pollPendingQuery(this._conn));
                 } catch (e: any) {
                     console.log(e);
-                    reject(e);
+                    // If the worker is not set, the worker has been terminated
+                    if (e.message.includes('worker is not set!')) {
+                        reject(new Error('Worker has been terminated'));
+                    } else {
+                        //Otherwise, reject with the error
+                        reject(e);
+                    }
+                    
                 }
             });
         }


carlopi:
Thanks!!

PR #1996, opened 2025-05-01
Make sure we `Fetch` query result when ready

Y--:
In the current code, we call `Fetch` on a query result before checking if there is a result available, which may cause deadlocks since we don't return to the event loop in the `BLOCKED` and `NO_TASKS_AVAILABLE` cases.

This PR solves it by returning a special buffer containing one "magic" byte out of the C++ layer, then a `null` buffer indicating the caller they need to retry fetching.

We tested this with MotherDuck operators which do return "BLOCKED", and this fixes the streaming issue we had.

Ideally, this would test this with a custom "test" extension that reproduces the behavior.

Another way to test it (albeit less ideal), would be to run DuckDB Wasm tests with a special version of DuckDB compiled with `FORCE_ASYNC_SINK_SOURCE` flag.

### DIFF
--- a/packages/duckdb-wasm/src/bindings/bindings_base.ts
+++ b/packages/duckdb-wasm/src/bindings/bindings_base.ts
@@ -4,7 +4,7 @@ import { Logger } from '../log';
 import { InstantiationProgress } from './progress';
 import { DuckDBBindings } from './bindings_interface';
 import { DuckDBConnection } from './connection';
-import { StatusCode, IsArrowBuffer } from '../status';
+import { StatusCode, IsArrowBuffer, IsDuckDBWasmRetry } from '../status';
 import { dropResponseBuffers, DuckDBRuntime, readString, callSRet, copyBuffer, DuckDBDataProtocol } from './runtime';
 import { CSVInsertOptions, JSONInsertOptions, ArrowInsertOptions } from './insert_options';
 import { ScriptTokens } from './tokens';
@@ -224,6 +224,11 @@ export abstract class DuckDBBindingsBase implements DuckDBBindings {
     /** Fetch query results */
     public fetchQueryResults(conn: number): Uint8Array | null {
         const [s, d, n] = callSRet(this.mod, 'duckdb_web_query_fetch_results', ['number'], [conn]);
+        if (IsDuckDBWasmRetry(s)) {
+            dropResponseBuffers(this.mod);
+            return null; // Retry
+        }
+
         if (!IsArrowBuffer(s)) {
             throw new Error("Unexpected StatusCode from duckdb_web_query_fetch_results (" + s + ") and with self reported error as" + readString(this.mod, d, n));
         }
@@ -233,11 +238,6 @@ export abstract class DuckDBBindingsBase implements DuckDBBindings {
 
         const res = copyBuffer(this.mod, d, n);
         dropResponseBuffers(this.mod);
-
-        // Special case indicating to the caller they need to retry
-        if (res.length === 1 && res[0] == 84) {
-            return null;
-        }
         return res;
     }
     /** Get table names */
--- a/packages/duckdb-wasm/src/bindings/bindings_base.ts
+++ b/packages/duckdb-wasm/src/bindings/bindings_base.ts
@@ -222,7 +222,7 @@ export abstract class DuckDBBindingsBase implements DuckDBBindings {
         return this.mod.ccall('duckdb_web_pending_query_cancel', 'boolean', ['number'], [conn]);
     }
     /** Fetch query results */
-    public fetchQueryResults(conn: number): Uint8Array {
+    public fetchQueryResults(conn: number): Uint8Array | null {
         const [s, d, n] = callSRet(this.mod, 'duckdb_web_query_fetch_results', ['number'], [conn]);
         if (!IsArrowBuffer(s)) {
             throw new Error("Unexpected StatusCode from duckdb_web_query_fetch_results (" + s + ") and with self reported error as" + readString(this.mod, d, n));
@@ -230,8 +230,14 @@ export abstract class DuckDBBindingsBase implements DuckDBBindings {
         if (s !== StatusCode.SUCCESS) {
             throw new Error(readString(this.mod, d, n));
         }
+
         const res = copyBuffer(this.mod, d, n);
         dropResponseBuffers(this.mod);
+
+        // Special case indicating to the caller they need to retry
+        if (res.length === 1 && res[0] == 84) {
+            return null;
+        }
         return res;
     }
     /** Get table names */
--- a/packages/duckdb-wasm/src/parallel/async_connection.ts
+++ b/packages/duckdb-wasm/src/parallel/async_connection.ts
@@ -115,7 +115,7 @@ export class AsyncResultStreamIterator implements AsyncIterable<Uint8Array> {
     /** Reached end of stream? */
     protected _depleted: boolean;
     /** In-flight */
-    protected _inFlight: Promise<Uint8Array> | null;
+    protected _inFlight: Promise<Uint8Array | null> | null;
 
     constructor(
         protected readonly db: AsyncDuckDB,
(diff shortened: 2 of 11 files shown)


packages/duckdb-wasm/src/bindings/bindings_base.ts
```diff
@@ -222,13 +222,19 @@ export abstract class DuckDBBindingsBase implements DuckDBBindings {
...
     /** Fetch query results */
-    public fetchQueryResults(conn: number): Uint8Array {
+    public fetchQueryResults(conn: number): Uint8Array | null {
         const [s, d, n] = callSRet(this.mod, 'duckdb_web_query_fetch_results', ['number'], [conn]);
         if (s !== StatusCode.SUCCESS) {
             throw new Error(readString(this.mod, d, n));
         }
+
         const res = copyBuffer(this.mod, d, n);
         dropResponseBuffers(this.mod);
```
carlopi, line 232:
```suggestion
        dropResponseBuffers(this.mod);
```
I am not sure if this needs to be performed, or moved afterwards.

packages/duckdb-wasm/src/parallel/async_connection.ts
```diff
@@ -115,7 +115,7 @@ export class AsyncResultStreamIterator implements AsyncIterable<Uint8Array> {
     /** Reached end of stream? */
     protected _depleted: boolean;
     /** In-flight */
-    protected _inFlight: Promise<Uint8Array> | null;
+    protected _inFlight: Promise<Uint8Array | null> | null;
```
carlopi, line 118: This change is probably not needed, given buffer has to be valid on returning.

packages/duckdb-wasm/src/parallel/async_connection.ts
```diff
@@ -135,17 +135,21 @@ export class AsyncResultStreamIterator implements AsyncIterable<Uint8Array> {
...
-        } else {
+        }
+
+        while (buffer == null) {
             buffer = await this.db.fetchQueryResults(this.conn);
         }
+
         this._depleted = buffer.length == 0;
         if (!this._depleted) {
             this._inFlight = this.db.fetchQueryResults(this.conn);
```
carlopi, line 150: Should also this be a while? Unsure.

packages/duckdb-wasm/src/bindings/bindings_base.ts
```diff
@@ -222,13 +222,19 @@ export abstract class DuckDBBindingsBase implements DuckDBBindings {
...
+
+        // Special case indicating to the caller they need to retry
+        if (res.length === 1 && res[0] == 84) {
+            return null;
+        }
```
carlopi, line 237: Alternative would be returning an arrow error with custom error code, unsure if that is properly supported, let me check.

carlopi (approved):
This is properly hacky, love it.

Left some comments, for you or for me in a bit.

I have a setup to run most duckdb unittester in duckdb-wasm (currently only node) vs various duckdb-wasm interfaces, I need to upstream that, and that can be used to stress test this more.

carlopi:
Actually not properly approved, left some comments that I need to double check.

packages/duckdb-wasm/src/bindings/bindings_base.ts
Y--: Yeah that's the other option we considered, and I agree it should work just as well. We felt this way was the most straightforward and explicit compared to abuse a specific Arrow error code.

packages/duckdb-wasm/src/parallel/async_connection.ts
Y--: This `fetchQueryResults` pro-actively request the engine to fetch the next chunk. Since we have the loop below we don't need to loop here.

packages/duckdb-wasm/src/parallel/async_connection.ts
Y--: I believe it is needed since `fetchQueryResults` may return `null` (when there's no chunk yet)

packages/duckdb-wasm/src/bindings/bindings_base.ts
Y--: Yeah it may need to be revisited though might be worth a dedicated PR?

packages/duckdb-wasm/src/parallel/async_connection.ts
carlopi: Yes, connected to next comment, I guess if the answer there is no while, here point is moot.

packages/duckdb-wasm/src/parallel/async_connection.ts
carlopi: Thanks

packages/duckdb-wasm/src/bindings/bindings_base.ts
carlopi: I was thinking moving this below the `return null` might be it, but I need to check.
In any case, this impact only `retry` cases, that are currently not exposed in the wild, so might be ok-ish.

Y--: I think we want it in both "SUCCESS" cases since a buffer is instantiated, otherwise we would leak the (1-byte) retry buffer each time no?

carlopi:
@Y--: I think the best path forward is wrapping Uint8Array in an object that can then hold also additional informations about what kind of buffer is returned and allows more flexibility, this is still needed going forward.

At that point instead of the magic byte a duckdb-wasm level error can be cleanly returned with a payload.

Y--:
> > @Y--: I think the best path forward is wrapping Uint8Array in an object that can then hold also additional informations about what kind of buffer is returned and allows more flexibility, this is still needed going forward.
> > At that point instead of the magic byte a duckdb-wasm level error can be cleanly returned with a payload.
> 
> I can try to work on that ~ my main concern is to do this preserving the 0-copy transfer as much as possible, but hopefully the Arrow API can help here.

Sadly it looks like [ConcatenateBuffers](https://arrow.apache.org/docs/cpp/api/memory.html#_CPPv418ConcatenateBuffersRK12BufferVectorP10MemoryPool) does [copy](https://github.com/apache/arrow/blob/692c8b7a597cf561c4788a7332d2681ed2002dd9/cpp/src/arrow/buffer.cc#L221) its input

carlopi:
I ended up sketching an alternative here: https://github.com/duckdb/duckdb-wasm/compare/main...carlopi:duckdb-wasm:refs/heads/wrap_fetch, what do you think? I might clean up and PR that, then you can add your on top

carlopi:
Hi, I ended up with https://github.com/duckdb/duckdb-wasm/pull/2000, after which it should be somewhat trivial to add this on top, it should be a matter of adding two enums, one at [lib/include/duckdb/web/webdb.h](https://github.com/duckdb/duckdb-wasm/pull/2000/files#diff-4e02e27230e28cc840005b8eef8909894cb611392e2ec929541bd1b9a8e59129) on the C++ side and one at [packages/duckdb-wasm/src/status.ts](https://github.com/duckdb/duckdb-wasm/pull/2000/files#diff-26ab371dc8c012bd599118c0c1db8f0a5ce5632ca08ebbe69ba1dd8ca70f8d62) in the typescript layer.

Plus adding the relevant handling in `fetchQueryResults` such as `return null;`

Would you mind taking a look, feedback is welcome

Y--:
> Hi, I ended up with #2000, after which it should be somewhat trivial to add this on top, it should be a matter of adding two enums, one at [lib/include/duckdb/web/webdb.h](https://github.com/duckdb/duckdb-wasm/pull/2000/files#diff-4e02e27230e28cc840005b8eef8909894cb611392e2ec929541bd1b9a8e59129) on the C++ side and one at [packages/duckdb-wasm/src/status.ts](https://github.com/duckdb/duckdb-wasm/pull/2000/files#diff-26ab371dc8c012bd599118c0c1db8f0a5ce5632ca08ebbe69ba1dd8ca70f8d62) in the typescript layer.
> 
> Plus adding the relevant handling in `fetchQueryResults` such as `return null;`
> 
> Would you mind taking a look, feedback is welcome

Had a quick look and it look very reasonable. I would need to look in more detail to make sure we don't copy the buffer un-necessarily, but I assume you already made sure of it?

carlopi:
> I would need to look in more detail to make sure we don't copy the buffer un-necessarily, but I assume you already made sure of it?

I will check this myself and verify this is the case. I would be inclined to say so, but the question is reasonable and better to get another look.

Y--:
Thanks for landing #2000 @carlopi ! I've rebased this PR on it, let me know if it needs more help :-)

carlopi (approved):
Thanks!

PR #2000, opened 2025-05-07
Wrap results of fetchQueryResults via DuckDBWasmResultsWrapper

carlopi:
This is currently just adding infrastructure, there should be no relevant changes

Note that enum added to lib/include/duckdb/web/webdb.h would evetntually need to be dupliciated also to packages/duckdb-wasm/src/status.ts


carlopi:
Before merging:
- [ ] double check no buffer copies are performed (context https://github.com/duckdb/duckdb-wasm/pull/1996#issuecomment-2858266399)

PR #2013, opened 2025-05-12
Pass down no core_extensions, and avoid setting BUILD_EXTENSIONS

carlopi:
This cuts by 10% Wasm binary sizes, adding a run-time dependency on usage of certain functions.


carlopi:
To be investigated test failures.

PR #2019, opened 2025-05-16
Publish v1.3+ under the next tag

hrl20:
This should publish build of main and `v` anything that's not v1.2-* under the `next` tag.


carlopi:
This is now outdated, but thanks!

PR #2021, opened 2025-05-16
Add config option for arrow_lossless_conversion

jraymakers:
Add the ability to set the `arrow_lossless_conversion` configuration parameter.

PR #2042, opened 2025-06-02
pass arrow_lossless_conversion config when generating arrow

jraymakers:
Recently, I [added support](https://github.com/duckdb/duckdb-wasm/pull/2021) for setting the config options `arrow_lossless_conversion`. However, this option was not passed through when generating Arrow buffers, so it had no effect on how DuckDB Wasm transmitted results. This PR corrects this by passing through the config option in the places where DuckDB Wasm generates Arrow buffers to transmit results.


jraymakers:
FYI @carlopi. This is a follow-up to https://github.com/duckdb/duckdb-wasm/pull/2021.

carlopi:
Thanks!

PR #2047, opened 2025-06-13
support PIVOT with send

jraymakers:
PIVOT is transformed into multiple statements by the parser. Since PendingQuery doesn't support multiple statements, this causes PIVOT statements to fail when used with `send`.

This change fixes this by using ExtractStatements to transform the text passed to WebDB::Connection::PendingQuery into possibly multiple parsed statements. Each one is run in turn; results are returned for the last one.

I added a unit test for this case. I did have to skip some of the checks (of the aggregated values) on Node, because they were returned as different types depending on the build (loadable or not). They work fine in the browser.

### DIFF
--- a/lib/include/duckdb/web/webdb.h
+++ b/lib/include/duckdb/web/webdb.h
@@ -57,6 +57,12 @@ class WebDB {
         /// The connection
         duckdb::Connection connection_;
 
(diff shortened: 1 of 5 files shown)


jraymakers:
@carlopi Finally got the tests to pass.

carlopi:
Thanks!

PR #2060, opened 2025-07-03
Add forceFullHTTPReads filesystem config flag.

jheer:
This PR adds a new filesystem configuration option: `forceFullHTTPReads`. If enabled, this flag suppresses the use of HTTP range requests for DuckDB-WASM is running in the browser.

The flag is set via the database `open` method:

```js
const worker = new Worker(config.mainWorker);
const db = new AsyncDuckDB(new ConsoleLogger(), worker);
await db.instantiate(config.mainModule, config.pthreadWorker);
await db.open({
  filesystem: {
    forceFullHTTPReads: true
  }
});
```

The flag passes in through WASM (C code), and back to a worker thread (TypeScript code) where the actual file loads are performed.

**Why add this flag?**

- Firefox has bad interactions with range requests and compression on some servers (including GitHub Pages), causing file loading in DuckDB-WASM to crash rather than gracefully fallback to full reads. Setting this flag sidesteps the error and enables successful file loading.
- When one knows ahead of time that the full data is needed, full file reads can be more efficient by making a single HTTP call rather than multiple.
- Full file reads result in fewer operations when connecting to object storage like R2 and S3, potentially reducing bills.

### DIFF
--- a/lib/include/duckdb/web/config.h
+++ b/lib/include/duckdb/web/config.h
@@ -71,6 +71,8 @@ struct DuckDBConfigOptions {
 struct FileSystemConfig {
     /// Allow falling back to full HTTP reads if the server does not support range requests
     std::optional<bool> allow_full_http_reads = std::nullopt;
+    /// Force full HTTP reads, suppressing use of range requests
+    std::optional<bool> force_full_http_reads = std::nullopt;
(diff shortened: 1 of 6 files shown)


Fil:
Thanks! Do we have a recommended pattern in the case where we don't know if we need to enable this flag? A try/catch?

PR #2093, opened 2025-09-07
Fix send('SELECT ...') infinite loop on detached state (no worker defined)

bluehat974:
This PR will prevent infinite loop in `connection.send('SELECT ...')` function, when _bindings is in detached mode (no worker associated)

There is a while loop waiting header variable to be defined. However,`this._bindings.pollPendingQuery(this._conn)` will just emit a console.error and return undefined, which will cause infinite loop.  

@carlopi if you have any improvement suggestion feel free.

Related issue: https://github.com/duckdb/duckdb-wasm/issues/1970


carlopi:
Thanks! Makes sense to me, let's see CI

PR #2123, opened 2025-11-05
Restore previous behaviour of getting full doc size from Content-Range

jbruggem:
Restores the previous behaviour of using the `Content-Range` header (when available) to extract the full content length from a partial request.

See:
* discussion here: https://github.com/duckdb/duckdb-wasm/pull/2057/commits/51b5481014d69fb587e1878088123731d2370cf5#r2490466628
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Range
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Length


jbruggem:
Fixed the linting issue. 

(I ran the linting script locally, this time, and it seemed fine)

shaarman:
See https://github.com/duckdb/duckdb-wasm/issues/2228

shaarman (approved):

jbruggem:
Hello,

now that this PR is approved, what's the process to get it merged ? 

thanks !

Issue #1900, opened 2024-10-27
Building from source fails in MacOS (it says missing JS files)


preddy4690:
### What happens?

I'm trying to build duckdb-wasm on my new MacOS. It gives an error when I run 'make serve' after running other commands. It can't find duckdb-mvp.js in ./src/bindings folder.

See attached screenshot.
![duckdb-wasm error](https://github.com/user-attachments/assets/e94d11fd-512c-461d-a3ef-a7541c1b09b9)

### To Reproduce

I ran these commands as instructed.

git clone https://github.com/duckdb/duckdb-wasm.git
cd duckdb-wasm
git submodule init
git submodule update
make apply_patches
make serve -- The error shows up here. 

### Browser/Environment:

Latest Chrome

### Device:

Macbook M3 Max

### DuckDB-Wasm Version:

1.29.0

### DuckDB-Wasm Deployment:

local

### Full Name:

Preetham Reddy

### Affiliation:

Tech Fabric

carlopi:
Can you possibly try adding:
```
make wasm_relsize
```
before `make serve` ?

I need to check why this has not been caught and fix the Makefile, this is very likely on me having changed the Makefile. Thanks.

neolord0:
I got the same results. 

"make serve" gave an error, but "make wasm_relsize" worked fine.

pgarrison:
Copied from #2139:

> I am trying to build duckdb-wasm from source, on macOS. I am following these instructions from the README:
> ```
> git clone https://github.com/duckdb/duckdb-wasm.git
> cd duckdb-wasm
> git submodule init
> git submodule update
> make apply_patches
> make serve
> ```
> 
> In addition to those steps, I've installed `ccache` (from homebrew) and `emsdk` (from source).
> 
> My build is failing at `make serve` with the following error:
> ```
> Error: Command failed: sed -i.bak 's/require(["'\'']child_process["'\''])/["child_process"].map(require)/g' ./src/bindings/duckdb-mvp.js && rm ./src/bindings/duckdb-mvp.js.bak
> sed: ./src/bindings/duckdb-mvp.js: No such file or directory
> 
>     at genericNodeError (node:internal/errors:984:15)
>     at wrappedFn (node:internal/errors:538:14)
>     at checkExecSyncError (node:child_process:891:11)
>     at execSync (node:child_process:963:15)
[...]

PR #2138, opened 2025-11-14
opfs: enable opfs temp_directory via pre-registered pool

### LINKED ISSUE #2061: Support for spilling to OPFS
### What happens?

Currently, setting 
```
SET temp_directory='opfs://temp'
```
causes this error when memory overflow occurs:
`Opening file 'opfs://temp/duckdb_temp_storage_S96K-0' failed with error: Error: No OPFS access handle registered with name: opfs://temp/duckdb_temp_storage_S96K-0`

### To Reproduce

N/a

### Browser/Environment:

Chrome

### Device:

Linux

### DuckDB-Wasm Version:

1.29.1-dev260.0

### DuckDB-Wasm Deployment:

N/a

### Full Name:

Debanjan Dhar

### Affiliation:

N/a

dt:
Previously attempting to set temp_directory to an `opfs://` path did not work. 

Specifically, while 'auto' file handling scanned SQL stmts for `opfs://` paths to pre-open mentioned files via the async OPFS API and register them where the sync openFile call could find them, it only worked for specific *files* and not for a temp *directory*, i.e. a path _in which_ `openFile` would later attempt to create individual files.

The sync vs async APIs of openFile vs OPFS presents a challenge here: when the DB decides to spill to a tempfile and attempts to create it, it does so via the `openFile` call, which is a sync call. This makes creating the OPFS file at that point, via the async OPFS API, a problem.

To get around this, as of this change, when an OPFS temp directory is registered, it sets up a 'pool' of pre-created empty temp files, with sync access handles to each pre-created, which can then be handed out as needed to (sync) `openFile` calls. When closed, a file can be truncated and returned to the pool, or be deleted if the pool is full. When the pool runs low on pre-created files, new files can be opened -- async -- and added to it.

This somewhat indirect approach works around the sync vs async API mismatch, while still allowing openFile calls to create arbitrarily named -- from its point of view -- files at arbitrary times. The included tests show this in action for opfs temp directories, both when configured via `SET` at interactively and via up-front configuration passed to `open()`.

Fixes #2061.

### DIFF
--- a/packages/duckdb-wasm/src/bindings/bindings_base.ts
+++ b/packages/duckdb-wasm/src/bindings/bindings_base.ts
@@ -13,6 +13,7 @@ import { arrowToSQLField, arrowToSQLType } from '../json_typedef';
 import { WebFile } from './web_file';
 import { UDFFunction, UDFFunctionDeclaration } from './udf_function';
 import * as arrow from 'apache-arrow';
+import { createOPFSTempPool } from '../utils/opfs_util';
 
 const TEXT_ENCODER = new TextEncoder();
 
@@ -701,6 +702,21 @@ export abstract class DuckDBBindingsBase implements DuckDBBindings {
             throw new Error('Not an OPFS file name: ' + file);
         }
     }
+
+    public async registerOPFSTempDir(tempPath?: string, maxPoolSize?: number, minPoolSize?: number): Promise<void> {
+        // Access BROWSER_RUNTIME through the runtime field
+        const runtime = this._runtime as any;
+
+        if (runtime._opfsTmpPool) {
+            await runtime._opfsTmpPool.destroy();
+            runtime._opfsTmpPool = null;
+        }
+
+        if (tempPath) {
+            runtime._opfsTmpPool = await createOPFSTempPool(tempPath, { maxUnused: maxPoolSize, minUnused: minPoolSize });
+        }
+    }
+
     public collectFileStatistics(file: string, enable: boolean): void {
         const [s, d, n] = callSRet(this.mod, 'duckdb_web_collect_file_stats', ['string', 'boolean'], [file, enable]);
         if (s !== StatusCode.SUCCESS) {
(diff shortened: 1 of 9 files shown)


pgarrison (approved):
I was able to use the artifacts from the feature branch's [github actions run](https://github.com/dt/duckdb-wasm/actions/runs/19406376443) to test this feature out in my application.

I did some manual testing with duckdb-wasm in my application, on Firefox. I used `db.open` with both `path` and `tempPath` pointing to the OPFS.
1. I set the same low memory limits as in the new tests to force spilling and ran the `spillQuery` from the new tests. I confirmed that opfs temp files were written.
2. Without low memory limits, I ran a query that fails on v1.30.0 because it runs out of memory, and it now succeeds. (I did this many times to evaluate performance for my use case; it was satisfactory.)

I was unable to build and test this, due to what seem to be macOS issues unrelated to the PR (#1900).

dt:
> unable to build

@pgarrison yeah, I hit this too on a fresh clone of `main`; I'm not certain but I think there are three distinct issues (all unrelated to this PR):

* `apply_patches` failed to apply for me; I ended up removing `no_httpfs.patch` and tweaking `preloaded_httpfs.patch` to get them to apply.
* The `Makefile` rule for `serve` seems to be missing a dependency, which I think is on `wasm_relsize`; you could run that by hand; I just added it as a dep to the `serve` rule but then I hit...
* it appears that newer versions of emscripten don't work, due to changes in how it handles pthreads and/or worker generation. Initially I thought anything before 3.1.58 would work, but in the end I needed to go down to Emscripten 3.1.50 to get it to work. I used `emscripten-core/emsdk` to initialize a local emscripten workspace env with the pinned version after which `make serve` succeeded for me.

Anyway, here's are diffs I had locally, which I think are everything I ended up doing to get `make serve` to work: https://github.com/duckdb/duckdb-wasm/compare/main...dt:duckdb-wasm:build

pgarrison:
Hey @dt, curious whether you closed this for lack of response or technical issues; I was looking forward to using it and could consider taking it over.

dt:
No technical issues that I know of, just seemed like upstream wasn't interested since it'd been open for couple months without any responses.

didimitrie:
Just adding our voice from Speckle that we'd be happy too!

PR #2146, opened 2025-12-09
Fix s3 endpoint paths

gregnr:
When the `s3_endpoint` URL contains a path (e.g. `https://s3.example.com/some/path`) and S3 path-style access is used, s3 requests fail with a `403 Forbidden` due to an incorrect AWS `AWS4-HMAC-SHA256` signature.

With the above example endpoint, the current logic will:
1. set `host` to  `s3.example.com/some/path`
2. set `path` `/<bucket-name>/<requested-path>`

This PR corrects this to:
1. set `host` to  `s3.example.com`
2. set `path` `/some/path/<bucket-name>/<requested-path>`

Which fixes the downstream signature that is generated from those values.

---

An example of an S3-compatible endpoint that does this in practice is Supabase's [Storage API](https://supabase.com/docs/guides/storage/s3/authentication), which serves the s3 endpoint from:
```
https://<project-ref>.storage.supabase.co/storage/v1/s3
```

### DIFF
--- a/packages/duckdb-wasm/src/utils/s3_helper.ts
+++ b/packages/duckdb-wasm/src/utils/s3_helper.ts
@@ -20,18 +20,9 @@ export interface S3PayloadParams {
     contentType: string | null;
 }
 
-const getEndpointPathPrefix = function (config: S3Config | undefined): string {
-    if (config?.endpoint?.startsWith('http')) {
-        const endpointUrl = new URL(config.endpoint);
-        // Return the path prefix if present, otherwise empty string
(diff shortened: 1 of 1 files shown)


gregnr:
@carlopi just a friendly ping on this. Anything I can do to help move it forward?

Side note: I attempted to add unit tests for the S3 path logic but struggled to fit it into the existing test frameworks (most other tests seemed to be full integration vs unit). Let me know if you'd like this in there though!

carlopi:
Thanks for the PR, LGTM!

PR #2197, opened 2026-03-25
Re-enable 'Load Existing DB File' OPFS test after fixes

Darker:
Two issues that I found:

First, when trying to use local build in my project, filename was being corrupted into `opfs:/xxx` with single slash. This is due to changes in duckdb main repo in filesystem.cpp, where it attempts to remove duplicate slashes.

Second, there was a missing registration for `.wal.checkpoint` files which duckdb now uses.

Tests pass after these two fixes.


Darker:
@carlopi This is in relation to PR #2184

PR #2198, opened 2026-03-25
Change build config to be compatible with emscripten 3x

Darker:
This is a draft, there may be more changes needed. I did these changes for my own use, but I wanted to share in case someone else is considering updating from emscripten <= 3.0

So far, following was changed:

 - Bump browser versions slightly to allow for bigint literals that emscriptenn now emits
 - HEAPxxx now needs to be explicitly exported
 - An additional js file is no longer emitted for pthread workers, instead emscripten inlines it into the main worker

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -296,7 +296,7 @@ if(EMSCRIPTEN)
       -s MAXIMUM_MEMORY=4GB \
       -s MODULARIZE=1 \
       -s EXPORT_NAME='DuckDB' \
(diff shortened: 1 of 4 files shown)


rustyconover:
I've been working on changes for Emscripten 5.x here 

https://github.com/Query-farm/duckdb-wasm/tree/emsdk-5.0.4-upgrade

Darker:
@rustyconover If you are already working on it, I am fine closing this PR. If any of my changes are useful, feel free to grab them and close this.

I primarily published it to allow others to use newer emscripten without doing thededdugging I had to do.

PR #2213, opened 2026-04-13
Add experimental wire api to duckdb-wasm

carlopi:
WIP


am2222:
@carlopi may i ask if this is related to this bug? I was not able to find a lot of documentations about `wire mode` 
https://github.com/duckdb/duckdb-wasm/issues/2216

PR #2232, opened 2026-06-12
Allow opening and dropping of OPFS databases

ThijsDreef:
# Allow opening and dropping of OPFS databases

## Summary

This PR makes it possible to attach an OPFS-backed database (`opfs://...`), write to it, detach, and then cleanly drop the underlying files, so that another tab or another DuckDB Wasm instance can open them afterwards.

OPFS support has been the subject of many PRs. This one doesn't try to fix everything; it targets two specific goals:

1. **Allow OPFS files to be used.**
2. **Allow a full attach / detach cycle of an OPFS database.**

It explicitly does **not** address automatic file registration.

## Allow OPFS files to be used

Recent changes to how file paths are canonicalized caused the input `opfs://file.db` to be rewritten as `opfs:/file.db`, breaking OPFS paths.

**Fix:** add a special case in the `buffered_filesystem` canonicalize function that preserves the path when it starts with `opfs://`.

## Allow attach / detach cycle of an OPFS database

The goal is for `register files` → `attach` → `query DB` → `detach` → `drop files` to release all files. Getting there required three fixes:

### 1. Register all associated files

Registering just the DB path and the WAL isn't enough — the temporary files associated with the database must be registered as well.

### 2. Handle `moveFile` for OPFS

On detach/checkpoint, `storage_manager.cpp` moves `*.wal.checkpoint` to `*.wal`. Moving files isn't supported by the OPFS filesystem.

A proper implementation would delete the target, create a new file, and copy the source's contents into it. But Wasm can't run asynchronous functions here, so instead we assume the file is already registered and move the bytes directly from one OPFS file to another. This keeps both files open and registered in case they're needed again.

**Fix:** special-case `moveFile` for OPFS file handling.

### 3. Re-open file with correct flags

Even with the above, writes still failed with errors about writing to a read-only file. This appears to be because re-open swaps the file handle without setting the flags. Setting the flags fixes it.

**Known trade-off:** this fix breaks one HTTPFS test. I think that's acceptable — as a user, I wouldn't expect an error in that test. If anyone disagrees, let me know and we can look for an alternative solution.

### DIFF
--- a/lib/include/duckdb/web/io/buffered_filesystem.h
+++ b/lib/include/duckdb/web/io/buffered_filesystem.h
@@ -141,6 +141,9 @@ class BufferedFileSystem : public duckdb::FileSystem {
         return filesystem_.Glob(PatchFilenameOwned(path), opener);
     }
 
+    /// Overide canonicalize path to allow for OPFS:// path urls
+    string CanonicalizePath(const string &path_p, optional_ptr<FileOpener> opener = nullptr) override;
+
     /// Register subsystem
     void RegisterSubSystem(unique_ptr<FileSystem> sub_fs) override;
     /// Register subsystem
--- a/lib/src/io/buffered_filesystem.cc
+++ b/lib/src/io/buffered_filesystem.cc
@@ -6,6 +6,7 @@
 #include <vector>
 
 #include "duckdb/common/file_system.hpp"
+#include "duckdb/common/string_util.hpp"
 #include "duckdb/main/client_context.hpp"
 #include "duckdb/web/io/buffered_filesystem.h"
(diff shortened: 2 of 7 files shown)


carlopi (commented):
I am not sure why CI had not run. I closed and reopened the PR (sorry for the noisy) that should trigger CI.

Code review makes sense to me, this is provably only touching the currently not-functional `opfs` code path. It's hard to prove this behaves correctly, but I think it's an improvement and can be iterated on.

I'll have a look at running this, and come back to this, but for now no comments on my side. Thanks!

carlopi:
Thanks!

PR #2238, opened 2026-06-29
Azure Blob / ADLS Gen2 support (abfss/az → https+SAS) (read only)

HynekBlaha:
Hello, I was working with Lakekeeper Iceberg Catalog when I noticed the built-in duckdb-wasm does not support reading iceberg tables from azure blob storage, because of missing azure auth library in wasm.

I got inspired by https://github.com/duckdb/duckdb-wasm/discussions/2217 and took my chance at implementing this feature. It works. I tried to make the changes as minimal/straight-forward as possible, but I would highly appreciate if someone experienced would take a look at it and cleaned it up / suggested improvements.

Before:

https://github.com/user-attachments/assets/5b7fada3-3b93-442b-9589-afbe1bd2f2e3

After:

https://github.com/user-attachments/assets/cf18a093-a166-45cf-a080-92210efb1427


**What it does**                                                                                                                                                                                                      
                                                                                                                                                                                                                    
  Adds an AzureFileSystem to the wasm layer that can access Azure Blob / ADLS Gen2 paths (abfss://, abfs://, az://, azure://, wasbs://, wasb://) over HTTPS using a SAS token — without the native azure extension, which
  can't be compiled to wasm.                                                                                                                                                                                        
                                                                                                                                                                                                                    
  Mechanism (three parts):                                                                                                                                                                                          
  1. Registers an azure secret type (provider config: account_name, connection_string) so that CREATE SECRET ... TYPE azure — issued by the Iceberg extension from vended credentials — succeeds.                   
  2. Registers the filesystem as a VFS subsystem that claims the Azure URL schemes. This also stops DuckDB from trying to autoload the unavailable native azure extension (today this fails with "Extension 'azure' 
  is not available").                                                                                                                                                                                               
  3. On open: looks up the matching azure secret (scoped to the original abfss:// path), rewrites the URL to HTTPS preserving the endpoint host (abfss→dfs, wasbs→blob), appends the SAS token as the query string, 
  and re-opens it through the buffered VFS as a regular https:// file. The SAS authorizes via the query string, so there is no request signing (unlike S3).                                                         
                                                                                                                                                                                                                    
  **What it brings**                                                                                                                                                                                                    
                                                                                                                                                                                                                    
  - Lets you query Iceberg tables stored on ADLS Gen2 directly from the browser (e.g. via a Lakekeeper REST catalog). Today this fails at azure-extension autoload.                                                 
  - Reuses the existing wasm HTTPS range-read path — no Azure SDK, no new runtime dependencies
  - No request signing; reads go through the same buffered path as s3/https.                                                                                                                                        
                                                                                                                                                                                                                    
  **Why not port the Azure SDK to wasm?**                                                                                                                                                                               
                                                                                                                                                                                                                    
  The alternative would be compiling DuckDB's azure extension (Azure SDK for C++) to wasm. 
  - No wasm build of the Azure C++ SDK exists (official or community). Its default transport is libcurl/WinHTTP, which don't work in browser wasm. A port is feasible — the SDK has a pluggable HttpTransport, so   
  one could implement an emscripten-fetch transport and build azure-core + azure-storage-* with crypto backed by the already-present mbedtls — but it is substantial work with no upstream support.                 
  - It buys little for the browser case. A browser has no managed identity / interactive AAD flow, so authentication is effectively limited to SAS/tokens regardless of the SDK. For read access via a vended SAS,  
  the full SDK is overkill: a URL rewrite covers it at a fraction of the cost.                                                                                                                                      
  - The SDKs that do run in a browser/wasm context are the wrong layer for this: the JS SDK (@azure/storage-blob) can't be called from the C++ extension, and the Rust SDK targets server-side WASI, not the        
  browser.                                                                                                                                                                                                          
                                                                                                                                                                                                                    
  **Scope / limitations**                                                                                                                                                                                               
                                                                                                                                                                                                                    
  - Works only with vended credentials by the catalog (requires an azure secret whose connection_string carries a SharedAccessSignature). No AAD/managed-identity auth.                                                                                                              
  - ADLS Gen1 (adl://) is not supported (the service was retired in 2024).                                                                                                                                          
  - Glob / directory operations on Azure paths are not implemented (NotImplementedException) — Iceberg uses explicit file paths.                                                                                    
  - The browser must be able to reach the storage account: cross-origin access requires CORS on the storage account (allow the UI origin, GET/HEAD/OPTIONS, the Range request header, and expose                    
  Content-Range/Content-Length). In the Lakekeeper UI the catalog itself is same-origin, so only ADLS needs CORS.                                                                                                   
  - Directory-scoped SAS (sr=d, as vended by Lakekeeper) is only valid on the dfs endpoint — hence host preservation rather than forcing the blob endpoint.                                                         
                                                                                                                                                                                                                    
  Tests                                                                                                                                                                                                             
                                                                                                                                                                                                                    
  Unit tests for URL parsing, SAS extraction, and the rewrite. Verified end-to-end against a Lakekeeper REST catalog over a real ADLS Gen2 table.

### DIFF
--- a/actions/image/Dockerfile
+++ b/actions/image/Dockerfile
@@ -22,7 +22,11 @@ RUN apt-get update -qq \
     && export CHROME_BIN=/usr/bin/google-chrome \
     && rm -rf /var/lib/apt/lists/* /etc/apt/sources.list.d/google.list
 
-ARG EMSDK_VERSION="3.1.56"
+# Keep in sync with EMSCRIPTEN_VERSION in .github/workflows/main.yml. Older emsdk (e.g. 3.1.56)
+# emits per-export `var X = wasmExports["X"]` bindings — including getTempRet0/setTempRet0 — which
+# the export-slimming awk in scripts/wasm_build_lib.sh strips, breaking loadable (MAIN_MODULE)
+# builds (instantiation fails / i64 return values are truncated).
+ARG EMSDK_VERSION="3.1.71"
 RUN mkdir -p /opt/emsdk \
     && cd /opt/emsdk \
     && curl -SL https://github.com/emscripten-core/emsdk/archive/${EMSDK_VERSION}.tar.gz | tar -xz --strip-components=1 \
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -224,6 +224,7 @@ add_library(
   ${CMAKE_SOURCE_DIR}/src/functions/table_function_relation.cc
   ${CMAKE_SOURCE_DIR}/src/insert_options.cc
   ${CMAKE_SOURCE_DIR}/src/io/arrow_ifstream.cc
+  ${CMAKE_SOURCE_DIR}/src/io/azure_filesystem.cc
   ${CMAKE_SOURCE_DIR}/src/io/buffered_filesystem.cc
   ${CMAKE_SOURCE_DIR}/src/io/file_page_buffer.cc
   ${CMAKE_SOURCE_DIR}/src/io/file_stats.cc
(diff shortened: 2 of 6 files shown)


HynekBlaha:
Hi @carlopi, what do you think about this PR? Could you please take a look 👀 and share your thoughts?
I tried to make the PR as minimal as possible, but it's hard for me to judge.
I attached a screenshot of webpage with locally built wasm from branch - it does what I expect. But is it enough?
Any opinion highly appreciated! ❤️

lputnam-hedgineer:
@HynekBlaha appreciate you implementing this. Also waiting on this myself

HynekBlaha:
I successfully built duckdb-wasm 1.5.4 with this commit on it. I was able to include this duckdb-wasm into current version of lakekeeper and all the UI views work as expected.
Could you please take a look at it @carlopi? 🙏

PR #2249, opened 2026-08-04
Fix OPFS moveFile() corrupting files longer than 1MB

### LINKED ISSUE #2248: OPFS moveFile() corrupts files larger than 1MB
### What happens?

Calling `moveFile(foo, bar)` where both `foo` and `bar` are OPFS handles and `foo` is bigger than 1MB results in a destination file `bar` which is padded to a multiple of 1MB, where the trailing bytes are junk left over in the copy buffer from the previous chunk.

This appears to have been introduced pretty recently, in https://github.com/duckdb/duckdb-wasm/pull/2232. The bug happens because the entire intermediate buffer is always copied to the destination file, even if `read()` returned fewer than 1024 \* 1024 bytes. It does not happen for small files (<1MB) because of the buffer that's allocated is sized as `Math.min(size, 1024 * 1024)`, so copying the entire buffer in this case is correct.

I believe that the fix is simply to check the return value of `read()`, which indicates how many bytes it actually wrote into the buffer, and limit `write()` to that many bytes too. I'm working on a patch to do this and will open a PR as soon as I've figured out how to build and test it (this is my first time contributing to duckdb-wasm).

### To Reproduce

I discovered this while working on some code that uses GDAL in the browser to copy some data from DuckDB into a GeoJSON file, like this:

```js
const dir = await navigator.storage.getDirectory();
for (const filename of ['out.geojson', 'tmp_out.geojson']) {
  const handle = await dir.getFileHandle(n, { create: true });
  await db.registerFileHandle(filename, handle, 3 /* BROWSER_FSACCESS */, true);
}

// GDAL writes tmp_out.geojson then renames it onto out.geojson, which triggers moveFile()
await conn.query(`COPY (...) TO 'out.geojson' WITH (FORMAT GDAL, DRIVER 'GeoJSON')`);
```

After running this, `out.geojson` is padded to a whole number of MB, with garbage trailing bytes which are duplicated from an earlier location in the file.

### Browser/Environment:

Firefox 152 (aarch64)

### Device:

MacBook Air (Apple M4)

### DuckDB-Wasm Version:

1.33.1-dev64.0

### DuckDB-Wasm Deployment:

n/a

### Full Name:

Jake Low

### Affiliation:

OpenStreetMap US

jake-low:
Fixes #2248.

I factored the actual copy logic out into a helper so I could test it. The important part of the change is checking the return value from `read()` and making sure to only `write()` that many bytes too. `new Uint8Array(buffer, offset, len)` creates a view over the buffer, not a copy, so there should be no performance cost to this change.

I didn't see a PR template or contributing guidelines, so please let me know if there's anything I should do to clean up or improve this (I did run the linter and formatter over my changes).


ThijsDreef:
Hey I think the change looks good overall, but I think it would be nice to add a test to the `opfs.test.ts` file as it would test the full end to end of things. Also I think you're right in adding a guard for zero value reads as they can happen according to the spec (https://fs.spec.whatwg.org/#api-filesystemsyncaccesshandle-read)

I think you can use a simplified form of you're reproducer #2248 to add a test to just `COPY (SELECT range(200_000) AS i) TO 'out.csv' (USE_TMP_FILE true)` (expect a 1.5mb csv file). Which exercises the move path and gives you a place to test it end to end.

A side note there are no unit tests specifically for internal parts of the DuckDB Wasm package (I believe them all to be E2E/Integration tests) as they all reach through the user facing interface, so I'm not sure on @duckdb opinion on adding a unit test specifically for this feature.

other side note, sorry for introducing this bug 🙃

jake-low:
Thanks for the suggestion! I will look into adding an integration test. I ran the unit tests I added in a hacky way (loading a prebuilt wasm bundle into a browser environment and then running the tests via the browser console) because I didn't have any luck trying to build the entire project locally. But I will try again.

> other side note, sorry for introducing this bug 🙃

No worries 🙂 I didn't mean to imply any blame when I linked to your PR from the issue. I uncovered this after updating a project I'm working on from 1.33.1-dev53.0 to 1.33.1-dev64.0, so I figured the regression must have crept in pretty recently and went and looked at recent PRs to track it down. The `latest` tag on npm still points to 1.33.1-dev57.0 which I think predates your change, so I'm betting very few other users have encountered this issue.

ThijsDreef:
So I took the liberty of coming up with a reproducer in the form of an integration test.

To reproduce I had to create a new instance with directIO disabled and register the files (tmp_moved.csv, moved.csv) and run `COPY (SELECT i FROM range(200_000) t(i)) TO 'moved.csv' (USE_TMP_FILE true)` 

This faithfully recreates the bug with a 2MB csv file, while the expected size is `1288892` bytes.

